PR #55, opened 2021-04-09
Update data source docs, broken links and add "How to use this plugin" section

sylviamoss:
In the overview page:
Screenshot 2021-04-09 at 11 32 05

### DIFF
--- a/docs/builders/index.mdx
+++ b/docs/builders/index.mdx
@@ -13,29 +13,29 @@ multiple builders depending on the strategy you want to use to build the AMI.
 
 The Amazon plugin supports the following builders at the moment:
 
-- [amazon-ebs](/docs/builders/amazon/ebs.mdx) - Create EBS-backed AMIs by
+- [amazon-ebs](/docs/builders/amazon/ebs) - Create EBS-backed AMIs by
   launching a source AMI and re-packaging it into a new AMI after
   provisioning. If in doubt, use this builder, which is the easiest to get
   started with.
 
-- [amazon-instance](/docs/builders/amazon/instance.mdx) - Create
+- [amazon-instance](/docs/builders/amazon/instance) - Create
   instance-store AMIs by launching and provisioning a source instance, then
   rebundling it and uploading it to S3.
 
-- [amazon-chroot](/docs/builders/amazon/chroot.mdx) - Create EBS-backed AMIs
+- [amazon-chroot](/docs/builders/amazon/chroot) - Create EBS-backed AMIs
   from an existing EC2 instance by mounting the root device and using a
   [Chroot](https://en.wikipedia.org/wiki/Chroot) environment to provision
   that device. This is an **advanced builder and should not be used by
   newcomers**. However, it is also the fastest way to build an EBS-backed AMI
   since no new EC2 instance needs to be launched.
 
-- [amazon-ebssurrogate](/docs/builders/amazon/ebssurrogate.mdx) - Create EBS
+- [amazon-ebssurrogate](/docs/builders/amazon/ebssurrogate) - Create EBS
   -backed AMIs from scratch. Works similarly to the `chroot` builder but does
   not require running in AWS. This is an **advanced builder and should not be
   used by newcomers**.
 
 -> **Don't know which builder to use?** If in doubt, use the [amazon-ebs
-builder](/docs/builders/amazon/ebs.mdx). It is much easier to use and Amazon
+builder](/docs/builders/amazon/ebs). It is much easier to use and Amazon
 generally recommends EBS-backed images nowadays.
 
 # Amazon EBS Volume Builder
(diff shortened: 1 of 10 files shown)


docs/datasources/ami.mdx
```diff
@@ -61,3 +77,26 @@ This is most useful for selecting a daily distro build.
...
 - `tags` (map[string]string) - The key/value combination of the tags assigned to the AMI.
 
+## Authentication
+
+The Amazon Data Sources authentication works just like for the [Amazon Builders](/docs/builders/amazon). Both
+have the same authentication options, and you can refer to the [Amazon Builders authentication](/docs/builders/amazon#authentication)
+to learn the options to authenticate for data sources.
+
+-> **Note:** A data source will start and execute in your own authentication session. The authentication in the data source
+doesn't relate with the authentication on Amazon Builders.
```
nywilken, line 87:
```suggestion
-> **Note:** The authentication session started by a data source is separate from any authentication sessions started by an Amazon builder. Users are encouraged to use `variables` for defining and sharing configuration values between datasources and builders. 
```

docs/datasources/ami.mdx
```diff
@@ -61,3 +77,26 @@ This is most useful for selecting a daily distro build.
 
 - `tags` (map[string]string) - The key/value combination of the tags assigned to the AMI.
 
+## Authentication
+
+The Amazon Data Sources authentication works just like for the [Amazon Builders](/docs/builders/amazon). Both
+have the same authentication options, and you can refer to the [Amazon Builders authentication](/docs/builders/amazon#authentication)
+to learn the options to authenticate for data sources.
```
nywilken, line 84:
```suggestion
The authentication for Amazon Data Sources uses the same configuration options as [Amazon Builders](/docs/builders/amazon). To learn more about all of the available authentication options please see [Amazon Builders authentication](/docs/builders/amazon#authentication).
```

docs/datasources/ami.mdx
```diff
@@ -61,3 +77,26 @@ This is most useful for selecting a daily distro build.
 
 - `tags` (map[string]string) - The key/value combination of the tags assigned to the AMI.
 
+## Authentication
```
nywilken, line 80: This authentication option information is duplicated on the data source overview page https://www.packer.io/docs/datasources/amazon and on each of the data sources. I recommend just adding it to the overview page for the datasources and linking back to it from each of the datasource documentation pages. 

The idea being that the usage code should be the biggest thing on the screen for each plugin.

docs/builders/index.mdx
```diff
@@ -13,37 +13,53 @@ multiple builders depending on the strategy you want to use to build the AMI.
...
   -backed AMIs from scratch. Works similarly to the `chroot` builder but does
   not require running in AWS. This is an **advanced builder and should not be
   used by newcomers**.
 
 -> **Don't know which builder to use?** If in doubt, use the [amazon-ebs
-builder](/docs/builders/amazon/ebs.mdx). It is much easier to use and Amazon
+builder](/docs/builders/amazon/ebs). It is much easier to use and Amazon
 generally recommends EBS-backed images nowadays.
 
+## How to use this plugin
```
nywilken, line 41: Hmmm...part of me thinks that this information should only be on the overview page for each component type, as opposed to being on all documentation pages.

 I don't feel strongly about it to suggest a change but I find the duplication of the install instructions on each page  to be overwhelming.

nywilken (approved):
Overall looks good. I left a couple of suggestions and some feedback on the install code blocks. Feel free to merge when ready.

docs/datasources/ami.mdx
sylviamoss: I initially was planning on doing how you suggested but then I thought the person would be redirected to the data source overview to be then redirected to the builder overview. Do you think is a good idea remove it from the overview and left it in the data source page only?

docs/builders/index.mdx
sylviamoss: I also think is overwhelming 🤔  I added the info in the component page as well as the overview, for the same reason I added the authentication info to each of the data sources pages. 
When Megan was using the data sources she didn't think of looking in the overview page for more information, so I figure users will do the same. 
But even with that, I think I will remove it. It is a different situationn.

docs/datasources/ami.mdx
nywilken: > Do you think is a good idea remove it from the overview and left it in the data source page only?

I'm thinking the same way, and I see your comment below about users not checking the overview page. I think we generally go directly to the docs pages because we know it. So for new users over communicating may not be a bad thing. 

One thing I've seen in the Azure builder is that auth configuration was duplicated and then eventually drifted apart  so we need to make sure things stay up to date on all pages.

sylviamoss: I removed it from the overview 👀

docs/builders/index.mdx
sylviamoss: left it only in the overview page when there's one.

docs/datasources/ami.mdx
sylviamoss: Yes! At least both don't have much information and link to the builder's overview!

nywilken (approved):
LGTM

PR #78, opened 2021-05-11
Update import.mdx

besmirzanaj:
Update in the documentation to describe the need for `cloud-init`  in linux systems.

If package is not present, proper provisioning will fail for new EC2 instances.


sylviamoss (approved):
LGTM! 👍🏼

PR #82, opened 2021-05-26
Add Packer and packer-plugin-amazon to the User-Agent header

serverwentdown:
Resolves hashicorp/packer#10315, by adding the plugin version to the User-Agent header sent by the AWS SDK. This is code copied from [session.go in aws-sdk-go-base](https://github.com/hashicorp/aws-sdk-go-base/blob/da4108d88728fcf4f5f9df9ea817285bccf7078e/session.go#L105).

### DIFF
--- a/builder/common/access_config.go
+++ b/builder/common/access_config.go
@@ -20,6 +20,7 @@ import (
 	cleanhttp "github.com/hashicorp/go-cleanhttp"
 	"github.com/hashicorp/packer-plugin-amazon/builder/common/awserrors"
 	pluginversion "github.com/hashicorp/packer-plugin-amazon/version"
+	"github.com/hashicorp/packer-plugin-sdk/common"
 	vaultapi "github.com/hashicorp/vault/api"
 )
 
@@ -202,6 +203,10 @@ type AccessConfig struct {
 	PollingConfig *AWSPollingConfig `mapstructure:"aws_polling" required:"false"`
 
 	getEC2Connection func() ec2iface.EC2API
+
+	// packerConfig is set by Prepare() containing information about Packer,
+	// including the CorePackerVersionString
+	packerConfig *common.PackerConfig
 }
 
 // Config returns a valid aws.Config object for access to AWS services, or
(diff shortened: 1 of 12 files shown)


builder/common/access_config.go
```diff
@@ -281,6 +288,18 @@ func (c *AccessConfig) Session() (*session.Session, error) {
...
+	userAgentProducts := []*awsbase.UserAgentProduct{
+		{Name: "APN", Version: "1.0"},
+		{Name: "HashiCorp", Version: "1.0"},
+		{Name: "Packer", Version: c.packerConfig.PackerCoreVersion, Extra: []string{"+https://www.packer.io"}},
+		{Name: "packer-plugin-amazon", Version: pluginversion.Version, Extra: []string{"+https://www.packer.io/docs/builders/amazon"}},
+	}
+	for i := len(userAgentProducts) - 1; i >= 0; i-- {
+		product := userAgentProducts[i]
+		c.session.Handlers.Build.PushFront(request.MakeAddToUserAgentHandler(product.Name, product.Version, product.Extra...))
+	}
```
serverwentdown, line 301: If this entire function relied on `awsbase` instead of reimplementing [`GetSession`](https://github.com/hashicorp/aws-sdk-go-base/blob/da4108d88728fcf4f5f9df9ea817285bccf7078e/session.go#L74) (like [this](https://github.com/hashicorp/terraform-provider-aws/blob/bc48c2178fff1c5b1b142481069e45fcb5870bd5/aws/config.go#L386)), I think it could be much cleaner.

SwampDragons (changes requested):
Hi, thanks for this PR! The change itself looks good. 

Looks like this new repo doesn't have our test runner properly configured to test branches; that's my fault and I'll fix it now. 

That said, you can run tests locally by changing directores into your packer-plugin-amazon repo directory and calling `make test`. There's a failure because the tests don't expect any input to the `Prepare()` method you've modified: 

```
# github.com/hashicorp/packer-plugin-amazon/builder/common [github.com/hashicorp/packer-plugin-amazon/builder/common.test]
builder/common/access_config_test.go:40:21: not enough arguments in call to c.Prepare
```

Can you take a look and fix those?

SwampDragons:
Awesome, thanks for the fast turnaround!

SwampDragons (approved):

serverwentdown:
Thanks, no problem!

PR #83, opened 2021-05-26
Added support for fleet_tags / fleet_tag

siliconsheep:
Hey folks!

In our organisation, we have an enforced policy where engineers need to tag resources at request time (when supported) to facilitate cost allocation / team. An EC2 Fleet is such a resource, but Packer currently does send any TagSpecifications in the CreateFleetInput.

This PR fixes that by adding a "fleet_tags" and "fleet_tag" parameter to the various builders, which will send the requested tags with the fleet request.

Tests have been updated as well and seems to be working well in our environment.

Cheers!


builder/common/run_config.go
```diff
@@ -606,6 +614,13 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		errs = append(errs, fmt.Errorf("either instance_type or "+
 			"spot_instance_types must be specified, not both"))
 	}
+	
+	if c.FleetTags != nil {
+		if c.SpotPrice == "" || c.SpotPrice == "0" {
+			errs = append(errs, fmt.Errorf(
+				"fleet_tags should not be set when not requesting a spot instance"))
```
SwampDragons, line 621: Why is this? The SpotPrice is used inside of the fleet request

SwampDragons (approved):
This looks great to me -- I just have one question inline where I'm pretty sure the issue is my understanding of spot instances/fleets, not your code.

builder/common/run_config.go
siliconsheep: This check is there as FleetTags wouldn't be used if we're not requesting a spot fleet. To check whether we are requesting a spot fleet or not, we check the SpotPrice value, which is required in that case. This is actually copied from an already existing check for SpotTags, a few lines down:

```
if c.SpotTags != nil { 
    if c.SpotPrice == "" || c.SpotPrice == "0" {
        errs = append(errs, fmt.Errorf(
            "spot_tags should not be set when not requesting a spot instance"))
     }
}
```

SwampDragons: Oh! I was reading the conditional backwards. I thought you were having validation fail if the user _is_ using a spot price. Yes, I think this is fine.

PR #84, opened 2021-05-26
Add circle config for branch tests

SwampDragons:
need to be able to run branch tests on this repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,112 @@
+orbs:
+  win: circleci/windows@1.0.0
+
+version: 2.1
+
+executors:
+  golang:
+    docker:
+      - image: docker.mirror.hashicorp.services/circleci/golang:1.16
+    resource_class: medium+
+  darwin:
+    macos:
+      xcode: "12.0.0"
+
+commands:
+  install-go-run-tests-unix:
+    parameters:
+      GOOS:
+        type: string
+      GOVERSION:
+        type: string
+      HOME:
+        type: string
+        default: "~"
+    steps:
+      - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,112 @@
...
+        type: string
+    steps:
+      - checkout
+      - run: GOOS=<< parameters.GOOS >> GOARCH=<<parameters.GOARCH>> go build -ldflags="-s -w -X github.com/hashicorp/packer/version.GitCommit=${CIRCLE_SHA1}" -o ./pkg/packer_<< parameters.GOOS >>_<< parameters.GOARCH >> .
+      - run: zip ./pkg/packer_<< parameters.GOOS >>_<< parameters.GOARCH >>.zip ./pkg/packer_<< parameters.GOOS >>_<< parameters.GOARCH >>
+      - run: rm ./pkg/packer_<< parameters.GOOS >>_<< parameters.GOARCH >>
+      - persist_to_workspace:
+          root: .
+          paths:
+            - ./pkg/
```
sylviamoss, line 53: This is very packer-specific. Plugins don't have the GitCommit commit constant, we will need to implement that.  I also think the zip should have the plugin name.

sylviamoss (changes requested):
This looks good! 
I'm requesting changes because of the persistent part, I think it has to be adapted for plugins. Also, I would add this to our packer-repositories to add it to all of the plugins 🤔  what do you think?

sylviamoss:
Another option is to update the pull request action to also run for branches with something like 
```
on: [push, pull_request]
```

nywilken:
@sylviamoss these are some great suggestions. We merged a PR last week into packer-repositories to push this change to all repos. Would you mind making your suggested changes to the packer-repositories since you have these top of mind so that we can push the final config to all repos at once. I will approve, push, and apply. 

I'm closing this PR as it has been superseded by the packer-repositories change.

PR #86, opened 2021-05-31
goreleaser - point to the correct version package

sylviamoss:
https://github.com/hashicorp/packer-plugin-amazon/pull/82 breaks the versioning automation and this is the fix.


serverwentdown:
My bad, :joy:

serverwentdown (approved):

PR #87, opened 2021-06-01
Add and test circleci

sylviamoss:
I want to merge this and test if the configuration works before committing it to all of our plugins.


.circleci/config.yml
```diff
@@ -0,0 +1,112 @@
...
+  test-windows:
+    executor:
+      name: win/vs2019
+      shell: bash.exe
+    steps:
+      - install-go-run-tests-windows:
+          GOVERSION: "1.16"
+  check-lint:
+    executor: golang
+    resource_class: xlarge
```
nywilken, line 82: We can probably drop this down now that it is just the plugin code. But that is just me thinking out loud.

sylviamoss: I think you're right!

nywilken (approved):
LGTM :+1: :+1:

PR #88, opened 2021-06-01
Fix circleci job and generate missing code

sylviamoss:


version/version.go
```diff
@@ -1,5 +1,6 @@
 package version
 
+// TODO: remove this comment
```
azr, line 3: yes

azr (approved):
This is the best, you rock ! Now, to do the same with other plugins :shipit:

version/version.go
sylviamoss: hahaha got catch, I did this just to push a commit and forgot about it

PR #104, opened 2021-07-02
[WIP] Add snapshot import to amazon-import

### LINKED ISSUE #103: amazon-import -> Import as snapshot?
#### Community Note

#### Description

Instead of only having the possibility of importing the artifacts as an amazon AMI, which has a lot of requirements, we should be able to import the artifact as a snapshot, which has almost no requirements and its not even postprocessed by aws.

It could even have an extra step to register the AMI from that snapshot afterwards thus turning the snapshot into an AMI

#### Use Case(s)

Any OS that does not adhere to the requirements list from importing images[0] could benefit from this. In our case this arise from building https://github.com/rancher-sandbox/cOS-toolkit and trying to use packer while building our artifacts for the different cloud providers. We tried to use amazon-import to upload the resulting OVA images but it kept failing until we used the import-snapshot method.

[0] https://docs.aws.amazon.com/vm-import/latest/userguide/vmie_prereqs.html

#### Potential configuration

```diff
{
  "type": "amazon-import",
  "access_key": "YOUR KEY HERE",
  "secret_key": "YOUR SECRET KEY HERE",
  "region": "us-east-1",
  "s3_bucket_name": "importbucket",
  "license_type": "BYOL",
+"snapshot": true,
+"register_snapshot": true,
  "tags": {
    "Description": "packer amazon-import {{timestamp}}"
  }
}

```

#### Potential References
https://docs.aws.amazon.com/vm-import/latest/userguide/vmimport-import-snapshot.html
https://docs.aws.amazon.com/cli/latest/reference/ec2/register-image.html

Itxaka:
This patch adds the possibility of importing and image to amazon by
using the import snapshot api, which has lower requirements than the
image import.

It resuses the current post-process method but diverges once we need to
import the image. The image import is untouched, but the new snapshot
import does something similar to image. It uploads the artifact to s3,
then imports it as snapshot and finally registers the snapshot as a new
ami.

Closes #103 

Signed-off-by: Itxaka <igarcia@suse.com>

### DIFF
--- a/builder/common/state.go
+++ b/builder/common/state.go
@@ -206,6 +206,18 @@ func (w *AWSPollingConfig) WaitUntilImageImported(ctx aws.Context, conn *ec2.EC2
 	return err
 }
 
+func (w *AWSPollingConfig) WaitUntilSnapshotImported(ctx aws.Context, conn *ec2.EC2, taskID string) error {
+	importInput := ec2.DescribeImportSnapshotTasksInput{
+		ImportTaskIds: []*string{&taskID},
+	}
+
+	err := WaitForSnapshotToBeImported(conn,
+		ctx,
+		&importInput,
+		w.getWaiterOptions()...)
+	return err
+}
+
 // Custom waiters using AWS's request.Waiter
 
 func WaitForVolumeToBeAttached(c *ec2.EC2, ctx aws.Context, input *ec2.DescribeVolumesInput, opts ...request.WaiterOption) error {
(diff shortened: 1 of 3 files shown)


Itxaka:
Missing:

 - [ ] docs
 - [ ] testing the import image manually to see if I broke it (probably!)
 - [ ] adding some extra options for the snapshot part (i.e. Architecture, boot mode, virtualization type)

Done:

 - [x] checked that I can generate an AMI from an artifact by going the snapshot route

Itxaka:
@sylviamoss Just the first pass, before I go further, I think this is a good step. Will await any comments before continuing. Have a nice weekend!

sylviamoss (commented):
This is looking good so far! I make some small suggestions. I haven't tested it though. 

So, my understanding is that importing from a snapshot is an easier option to register an AMI as it has fewer requirements to do so, right? In the end, the post-processor will produce two artifacts then, a snapshot and an AMI.
We need to make sure to document the advantages here, so people won't get confused about which option they should go for it.

Itxaka:
Thanks for the comments! Not stale, Im just a bit strapped on time :)

Itxaka:
Currently leaving for vacation, will pick this up once Im back in August!

sylviamoss:
Sounds good!

breisig:
Why was this never merged? This feature is required especially when attempting to upload Linux distro that aren't supported during vmimport [Example: Rocky Linux 9]

Itxaka:
See above, my bandwidth to test and continue developing this was zero, so I could not continue on it.

Good thing the code is in there, feel free to pick it up and with the comments from Silvia send a PR to have this merged 😀

PR #109, opened 2021-07-20
add license-specifications and placement/host-resource-group-arn

KhrisRichardson-BO:
This PR adds support for scheduling instances to [host resource groups](https://docs.aws.amazon.com/license-manager/latest/userguide/host-resource-groups.html), which likewise requires the declaration of a [license configuration](https://docs.aws.amazon.com/license-manager/latest/userguide/license-configurations.html).

While the AWS SDK allows for a collection of license configurations, this PR flattens that to one in accordance with some of the other conventions observed. I'm open to changing that decision, updating the data model to more accurately reflect the collection of license configurations that comprise a license specification.

In all honesty I would prefer to group the `tenancy` and `host_resource_group_arn` settings under a `placement` grouping, but I wasn't sure what the policy was around backwards compatibility or the general attitude around adherence to the AWS SDK schema vs. flattening to provoke the end user to use simple variable names that they can override via `vars` arguments on the command line.

Any input is welcome. Thanks

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -1,5 +1,5 @@
 //go:generate packer-sdc struct-markdown
-//go:generate packer-sdc mapstructure-to-hcl2 -type AmiFilterOptions,SecurityGroupFilterOptions,SubnetFilterOptions,VpcFilterOptions,PolicyDocument,Statement,MetadataOptions
+//go:generate packer-sdc mapstructure-to-hcl2 -type AmiFilterOptions,SecurityGroupFilterOptions,SubnetFilterOptions,VpcFilterOptions,PolicyDocument,Statement,MetadataOptions,LicenseConfigurationRequest,LicenseSpecification,Placement
 
 package common
 
@@ -42,6 +42,27 @@ type PolicyDocument struct {
 
 type SecurityGroupFilterOptions struct {
 	config.NameValueFilter `mapstructure:",squash"`
+}
+
+type LicenseConfigurationRequest struct {
+	// The Amazon Resource Name (ARN) of the license configuration.
+	LicenseConfigurationArn string `mapstructure:"license_configuration_arn"`
+}
+
+type LicenseSpecification struct {
+	// Describes a license configuration.
+	LicenseConfigurationRequest LicenseConfigurationRequest `mapstructure:"license_configuration_request"`
+}
+
+type Placement struct {
+	// The ARN of the host resource group in which to launch the instances.
+	HostResourceGroupArn string `mapstructure:"host_resource_group_arn" required:"false"`
+	// [Tenancy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-instance.html) used
+	// when Packer launches the EC2 instance, allowing it to be launched on dedicated hardware.
+	//
+	// The default is "default", meaning shared tenancy. Allowed values are "default",
+	// "dedicated" and "host".
+	Tenancy string `mapstructure:"tenancy" required:"false"`
 }
 
 // Configures the metadata options.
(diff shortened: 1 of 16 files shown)


azr:
> but I wasn't sure what the policy was around backwards compatibility or the general attitude around adherence to the AWS SDK schema vs. flattening

So the closer we are to the aws sdk schema and naming the better we generally think. In terms of backwards compatibility, a pattern we have often used when justified — and I think it would be justified here — was to have a deprecation period where both fields work in between.

Let me know what you prefer 🙂

KhrisRichardson-BO:
@azr I've updated the PR so that there's more coherence with the AWS SDK, including adding tenancy under placement with some deprecation warnings. I'm open to any suggestions.

P.S. there was one small ancillary edit made to make golangci-lint happy.

Thanks!

builder/common/run_config.go
```diff
@@ -390,11 +411,11 @@ type RunConfig struct {
...
 	SubnetId string `mapstructure:"subnet_id" required:"false"`
-	// [Tenancy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-instance.html) used
-	// when Packer launches the EC2 instance, allowing it to be launched on dedicated hardware.
-	//
-	// The default is "default", meaning shared tenancy. Allowed values are "default",
-	// "dedicated" and "host".
+	// [LicenseSpecification](#license-specification)
+	LicenseSpecifications []LicenseSpecification `mapstructure:"license_specifications" required:"false"`
+	// [Placement](#placement)
+	Placement Placement `mapstructure:"placement" required:"false"`
```
azr, line 417: I think it could be nice to add a little code snippet here to show how to use it.

builder/common/run_config.go
```diff
@@ -680,11 +701,21 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
-	if c.Tenancy != "" &&
-		c.Tenancy != "default" &&
-		c.Tenancy != "dedicated" &&
-		c.Tenancy != "host" {
-		errs = append(errs, fmt.Errorf("Error: Unknown tenancy type %s", c.Tenancy))
+	var tenancy string
+	tenancies := []string{c.Placement.Tenancy, c.Tenancy}
```
azr, line 705: Here, I would fail if both are set. Otherwise, I think it could be nice to document the setting priority.

builder/common/step_run_source_instance.go
```diff
@@ -204,6 +206,22 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
...
+					LicenseConfigurationArn: aws.String(licenseConfigurationArn),
+				},
+			}
+			runOpts.LicenseSpecifications = append(runOpts.LicenseSpecifications, licenseSpecifications...)
+		}
+	}
+
+	if s.HostResourceGroupArn != "" {
+		runOpts.Placement.HostResourceGroupArn = aws.String(s.HostResourceGroupArn)
+	}
```
azr, line 223: 👍🏼

datasource/secretsmanager/data.go
```diff
@@ -125,7 +125,7 @@ func (d *Datasource) Execute() (cty.Value, error) {
 	output := DatasourceOutput{
 		Value:        value,
 		SecretString: aws.StringValue(secret.SecretString),
-		SecretBinary: fmt.Sprintf("%s", secret.SecretBinary),
+		SecretBinary: string(secret.SecretBinary),
```
azr, line 128: 👍🏼

azr (approved):
This looks good to me. I would polish docs a little more :). I'll be waiting a little more for that.

builder/common/run_config.go
```diff
@@ -390,11 +411,77 @@ type RunConfig struct {
...
+	//   }
+	// ]
+	// ```
+	//
+	//   Each `license_configuration_request` describes a license configuration,
+	//   the properties of which are:
+	//
+	//   - `license_configuration_arn` (string) - The Amazon Resource Name (ARN)
+	//     of the license configuration.
+	//
```
azr, line 448: 🤩  so nice !

builder/common/run_config.go
```diff
@@ -390,11 +411,77 @@ type RunConfig struct {
 	// subnet-12345def, where Packer will launch the EC2 instance. This field is
 	// required if you are using an non-default VPC.
 	SubnetId string `mapstructure:"subnet_id" required:"false"`
-	// [Tenancy](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-instance.html) used
-	// when Packer launches the EC2 instance, allowing it to be launched on dedicated hardware.
+	// The license configurations.
```
azr, line 414: Should we describe the default behavior too here ?

It used to be `The default is "default", meaning shared tenancy. Allowed values are "default", "dedicated" and "host".`

So we could say: `By default, when this setting is unspecified, shared tenancy will be selected.` or something

azr:
Thanks for addressing the docs changes, they looked great, I just have a tiny question/suggestion, and we should be good !

builder/common/run_config.go
```diff
@@ -390,11 +411,77 @@ type RunConfig struct {
...
+	// ```
+	//
+	//   Each `license_configuration_request` describes a license configuration,
+	//   the properties of which are:
+	//
+	//   - `license_configuration_arn` (string) - The Amazon Resource Name (ARN)
+	//     of the license configuration.
+	//
+	LicenseSpecifications []LicenseSpecification `mapstructure:"license_specifications" required:"false"`
+	// Describes the placement of an instance.
```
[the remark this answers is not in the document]
khrisrichardson: @azr the [docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Placement.html) don't really go into much detail about the default placement. My interpretation is that these are scheduling hints that EC2 uses to filter the host onto which the instance is scheduled, otherwise EC2 chooses at random using black box heuristics that are unknown to/unknowable by the end-user.

azr (approved):
This looks pretty good to me, thanks for answering my question, your time, etc !

PR #129, opened 2021-09-08
Add HCP Packer registry image metadata to builder artifacts

nywilken:


builder/common/artifact.go
```diff
@@ -133,16 +128,32 @@ func (a *Artifact) stateAtlasMetadata() interface{} {
...
 func (a *Artifact) stateHCPPackerRegistryMetadata() interface{} {
-	metadata := make([]hcpPackerRegistryImage, 0, len(a.Amis))
 
-	for region, imageId := range a.Amis {
-		metadata = append(metadata, hcpPackerRegistryImage{
+	f := func(k, v interface{}) (*registryimage.Image, error) {
+
+		region, ok := k.(string)
+		if !ok {
+			return nil, fmt.Errorf("unexpected value of key in Amis map")
```
SwampDragons, line 136:
```suggestion
			return nil, fmt.Errorf("unexpected type of key in Amis map")
```

SwampDragons: nitpick but you're doing a type assertion here.

builder/common/artifact.go
```diff
@@ -133,16 +128,32 @@ func (a *Artifact) stateAtlasMetadata() interface{} {
...
+		}
+		imageId, ok := v.(string)
+		if !ok {
+			return nil, fmt.Errorf("unexpected value for %q in Amis map", region)
```
SwampDragons, line 140:
```suggestion
			return nil, fmt.Errorf("unexpected type for %q in Amis map", region)
```

SwampDragons: if it's not a string, does %q work?

SwampDragons (approved):
This looks great!

builder/common/artifact.go
nywilken: Oh good catch, it wont work.

builder/common/artifact.go
```diff
@@ -133,16 +128,32 @@ func (a *Artifact) stateAtlasMetadata() interface{} {
...
-		metadata = append(metadata, hcpPackerRegistryImage{
+	f := func(k, v interface{}) (*registryimage.Image, error) {
+
+		region, ok := k.(string)
+		if !ok {
+			return nil, fmt.Errorf("unexpected type of key in Amis map")
+		}
+		imageId, ok := v.(string)
+		if !ok {
+			return nil, fmt.Errorf("unexpected type for %q in Amis map", region)
```
nywilken, line 140:
```suggestion
			return nil, fmt.Errorf("unexpected type for value in Amis map")
```

SwampDragons:
Merge when ready 🎉

azr (approved):

PR #161, opened 2021-11-11
Validate plugin from the packer-sdc plugin-validate command

azr:
* Use packer-sdc plugin-check from imported sdk in makefile 
* make sure we have the most recent sdk
* make goreleaser use that too


nywilken (approved):
LGTM

PR #173, opened 2022-01-07
Import post processor: Allow creation of AMIs with uefi boot & graviton support 

anish:
This PR allows creation of [UEFI based](https://aws.amazon.com/about-aws/whats-new/2021/03/amazon-ec2-now-supports-uefi-boot-when-migrating-virtual-machines-to-ec2/) amis as well arm64/graviton based AMIs

### DIFF
--- a/post-processor/import/post-processor.go
+++ b/post-processor/import/post-processor.go
@@ -48,6 +48,8 @@ type Config struct {
 	LicenseType     string            `mapstructure:"license_type"`
 	RoleName        string            `mapstructure:"role_name"`
 	Format          string            `mapstructure:"format"`
+	Architecture    string            `mapstructure:"architecture"`
+	BootMode        string            `mapstructure:"boot_mode"`
 
 	ctx interpolate.Context
 }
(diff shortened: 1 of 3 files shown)


post-processor/import/post-processor.go
```diff
@@ -224,6 +249,8 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packersdk.Ui, artifa
 				},
 			},
 		},
+		Architecture: &p.config.Architecture,
+		BootMode:     &p.config.BootMode,
```
azr, line 253: I wonder what happened when these were not passed before, do you know ? [Docs](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/ec2/import-image.html) are quite vague about it.

azr (approved):
This LGTM. Just had a question.

post-processor/import/post-processor.go
anish: I was wondering the same. It defaults to `x86_64` and `legacy-bios` if nothing is specified, but I could not quite pinpoint where

azr: Cool, thanks for checking ! These sound like the sensible defaults, yes.

PR #174, opened 2022-01-14
Parses filter option values as csv strings

### LINKED ISSUE #13: source_ami_filter permits only one value for name
_This issue was originally opened by @grodzik as hashicorp/packer#7295. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

Hi,

I'm building an AMI based on Debian base image (`debian-stretch-hvm-x86_64-gp2-*`). This requires doing `apt upgrade` during building, which slowly takes longer, taking into account that more updates come out since release of the given base image. Let's assume my image will be called `my-debian-based-image-{{timestamp}}`. I would like to give both names to filter out most recent. Kind of incremental build, only if last build is newer than `debian-stretch-hvm-x86_64-gp2-*`. Using AWS CLI I can filter images using both names, like so:

    aws ec2 describe-images --filters "Name=name,Values=debian-stretch-hvm-x86_64-gp2-*,my-debian-based-image-*" --query "Images[].Name"

Unfortuantely this is not possible to achieve in packers `source_ami_filter`. It's impossible to pass array for filters attributes. Passing whole Values string from above also doesn't work, which is errored in the console like so:

    2019/02/08 15:13:37 ui error: Build 'amazon-ebs' errored: No AMI was found matching filters: {
        Filters: [{
            Name: "name",
            Values: ["debian-stretch-hvm-x86_64-gp2-*,my-debian-based-image-*"]
        }]
    }

Looking at the source code (I'm not a Go programmer), I would say this has something to do with it: https://github.com/hashicorp/packer/blob/master/builder/amazon/common/build_filter.go#L13
where I would expect that string to be splited on `,` into array.

- packer version: 1.3.4
- platform: Debian Linux

lorengordon:
Currently, EC2 filter values are treated as a single string item. However, to the AWS API, they are lists of strings. `buildEc2Filters` creates the struct expected by the API, but does not handle the case where the user wants to pass several values for the filter option.

This patch parses the filter option values as csv strings, to maintain backwards compatibility with the current interface while allowing users to pass a multi-valued argument.

Closes #13

### DIFF
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -140,7 +140,8 @@ type Config struct {
 	//This will fail unless *exactly* one AMI is returned. In the above example,
 	//`most_recent` will cause this to succeed by selecting the newest image.
 	//
-	//-   `filters` (map of strings) - filters used to select a `source_ami`.
+	//-   `filters` (map{key string, value string | comma separated string}) -
+	//  filters used to select a `source_ami`.
 	//	NOTE: This will fail unless *exactly* one AMI is returned. Any filter
 	//	described in the docs for
 	//	[DescribeImages](http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html)
--- a/builder/common/build_filter.go
+++ b/builder/common/build_filter.go
@@ -24,7 +24,8 @@ func buildEc2Filters(input map[string]string) []*ec2.Filter {
 		}
 
 		for _, r := range values {
-			b = append(b, &r)
+			var value = r
+			b = append(b, &value)
 		}
 
 		filters = append(filters, &ec2.Filter{
--- a/builder/common/build_filter.go
+++ b/builder/common/build_filter.go
@@ -1,18 +1,35 @@
 package common
 
 import (
+	"encoding/csv"
+	"log"
+	"strings"
+
 	"github.com/aws/aws-sdk-go/service/ec2"
 )
 
 // Build a slice of EC2 (AMI/Subnet/VPC) filter options from the filters provided.
 func buildEc2Filters(input map[string]string) []*ec2.Filter {
 	var filters []*ec2.Filter
+
 	for k, v := range input {
+		var b []*string
+
 		a := k
-		b := v
+		csvReader := csv.NewReader(strings.NewReader(v))
+
+		values, err := csvReader.Read()
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		for _, r := range values {
+			b = append(b, &r)
+		}
+
 		filters = append(filters, &ec2.Filter{
 			Name:   &a,
-			Values: []*string{&b},
+			Values: b,
 		})
 	}
 	return filters
--- a/builder/common/build_filter.go
+++ b/builder/common/build_filter.go
@@ -17,6 +17,7 @@ func buildEc2Filters(input map[string]string) []*ec2.Filter {
 
 		a := k
 		csvReader := csv.NewReader(strings.NewReader(v))
+		csvReader.TrimLeadingSpace = true
 
 		values, err := csvReader.Read()
 		if err != nil {
--- a/builder/common/build_filter_test.go
+++ b/builder/common/build_filter_test.go
@@ -1,24 +1,83 @@
 package common
 
 import (
+	"strings"
 	"testing"
 )
 
-func TestStepSourceAmiInfo_BuildFilter(t *testing.T) {
+func TestStepSourceAmiInfo_BuildFilter_SingleValue(t *testing.T) {
 	filter_key := "name"
 	filter_value := "foo"
 	filter_key2 := "name2"
-	filter_value2 := "foo2"
+	filter_value2 := " foo2"
+	filter_key3 := "name3"
+	filter_value3 := "foo3 "
 
-	inputFilter := map[string]string{filter_key: filter_value, filter_key2: filter_value2}
+	inputFilter := map[string]string{
+		filter_key:  filter_value,
+		filter_key2: filter_value2,
+		filter_key3: filter_value3,
+	}
+
 	outputFilter := buildEc2Filters(inputFilter)
+
+	testFilter := map[string]string{
+		filter_key:  filter_value,
+		filter_key2: strings.TrimSpace(filter_value2),
+		filter_key3: filter_value3,
+	}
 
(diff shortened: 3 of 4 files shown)


lorengordon:
Hello, I just wanted to open this to get the discussion going. If this approach isn't acceptable, I'm happy to adjust it. I'll also certainly write tests once there is an accepted user interface. Thanks!

JenGoldstrich:
Hey @lorengordon thank you for submitting this PR! 

I manually built your PR and it seems to be working well to me, and seems to maintain backwards compatibility like you said with single strings, I think you can go ahead and add tests!

Linking AWS API docs that reference this filter field https://docs.aws.amazon.com/imagebuilder/latest/APIReference/API_Filter.html#imagebuilder-Type-Filter-values, which as you point out, accepts multiple filters.

I think with some testing this should be good to merge, thanks for this contribution @lorengordon

JenGoldstrich:
One additional thing also is that another GH user committed to this branch @dwc0011 who needs to sign the CLA above

JenGoldstrich (commented):
see above

lorengordon:
Great, thanks @JenGoldstrich! We'll get some tests and docs/examples in place shortly.

dwc0011:
@JenGoldstrich  I am working on creating the example and docs and wanted to make sure I understand things correctly.

Under the example directory, there is a single file build.pkr.hcl, is this file used/included/embedded anywhere in the documentation?  Do we need to add an additional example file for the list of name's or just adding to the documentation will suffice, since current example is still valid.  

If we add an additional example file is there a naming convention we should use? 

For the documentation, do we add to the partials and they are build into the docs?

JenGoldstrich:
I think y'all are probably fine without any docs change @dwc0011 , since we didn't specifically label the object type before hand anywhere but in the Packer docs here https://www.packer.io/plugins/builders/amazon/ebs#source_ami_filter,  I can make sure those docs get updated since I'm not 100% sure where they live.

JenGoldstrich:
Spoke too soon, I think all you would need to do is change https://github.com/hashicorp/packer-plugin-amazon/blob/main/docs-partials/builder/chroot/Config-not-required.mdx to indicate that the filters value can be a comma seperated string, to answer your earlier question yes that doc partial will be combined to create the Packer docsite

Instead of 
`filters (map of strings) - filters used to select a source_ami. NOTE: This will fail unless exactly one AMI is returned. Any filter described in the docs for DescribeImages is valid.`

You could put `filters (comma seperated string)...`

dwc0011:
> Spoke too soon, I think all you would need to do is change https://github.com/hashicorp/packer-plugin-amazon/blob/main/docs-partials/builder/chroot/Config-not-required.mdx to indicate that the filters value can be a comma seperated string, to answer your earlier question yes that doc partial will be combined to create the Packer docsite
> 
> Instead of `filters (map of strings) - filters used to select a source_ami. NOTE: This will fail unless exactly one AMI is returned. Any filter described in the docs for DescribeImages is valid.`
> 
> You could put `filters (comma seperated string)...`

Awesome, I will do tonight or first thing in the morning.  I appreciate the guidance!

dwc0011:
> Instead of `filters (map of strings) - filters used to select a source_ami. NOTE: This will fail unless exactly one AMI is returned. Any filter described in the docs for DescribeImages is valid.`
> 
> You could put `filters (comma seperated string)...`

After reviewing the documentation and the implementation, I slightly changed your suggestion.
In the end I went with 
```
filters` (map{key string, value string | comma separated string}) - 
filters used to select a `source_ami`.
```
As it is a map of key:value strings either single string or a csv string

If there is some better notation we can use, I am open to suggestion, I

lorengordon:
@JenGoldstrich I think we're good now! The patch works for me in my local build, both tests and docs have been updated, and CI is fully passing. Let us know if you need anything else!

builder/common/build_filter_test.go
```diff
@@ -29,3 +43,48 @@ func TestStepSourceAmiInfo_BuildFilter(t *testing.T) {
 		}
 	}
 }
+
+func TestStepSourceAmiInfo_BuildFilter_ListValue(t *testing.T) {
```
JenGoldstrich, line 47: Test looks good to me!

JenGoldstrich (approved):

JenGoldstrich:
Merged, thanks for this PR y'all!

lorengordon:
Thanks @JenGoldstrich!

dwc0011:
Awesome thank you very much!

builder/common/build_filter.go
```diff
@@ -1,18 +1,37 @@
...
+		var b []*string
+
 		a := k
-		b := v
+		csvReader := csv.NewReader(strings.NewReader(v))
+		csvReader.TrimLeadingSpace = true
+
+		values, err := csvReader.Read()
+		if err != nil {
+			log.Fatal(err)
```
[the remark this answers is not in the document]
nywilken: Good catch @azr. With that I would add that we generally don't tend to use log.Fatal or panic within the code base for the reasons mentioned by @azr. 

Bubbling up the error to the caller would be helpful here, and if the call is being made from a step where `ui` is avialable. Outputting the error information to the user with `Ui.Error(...)` would be nice to inform the user what happened when trying to build the filter.

builder/chroot/builder.go
```diff
@@ -140,7 +140,8 @@ type Config struct {
 	//This will fail unless *exactly* one AMI is returned. In the above example,
 	//`most_recent` will cause this to succeed by selecting the newest image.
 	//
-	//-   `filters` (map of strings) - filters used to select a `source_ami`.
+	//-   `filters` (map{key string, value string | comma separated string}) -
```
nywilken, line 143: Only suggestion here would be to make this `map[string]string` and as part of the comment specify that the value can be a comma separated list with an example.

builder/common/build_filter.go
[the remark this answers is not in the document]
dwc0011: For this one are you saying do Ui.Error(err) instead of log.Fatal(err)?  Just want to try and get it right without a bunch of commits.  Thanks!

builder/chroot/builder.go
dwc0011: I will look at this today, sorry I haven't addressed sooner

PR #179, opened 2022-01-20
Enable temporary_key_pair_type option for ed25519

### LINKED ISSUE #144: temporary_key_pair_type option has no effect
#### Overview of the Issue

Although [the documentation](https://www.packer.io/docs/builders/amazon/ebs#temporary_key_pair_type) lists it, the `temporary_key_pair_type` configuration option does not appear to actually do anything; furthermore, I do not see that option used at all in [builder/common/step_key_pair.go](https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/common/step_key_pair.go).

#### Reproduction Steps

1. Create a Packer Buildfile using the AWS EBS builder and include the option `"temporary_key_pair_type": "ed25519"`.  (I should mention that I am using a JSON template.)
2. Run `packer`
3. Observe in the AWS Console that the temporary EC2 keypair created is of type RSA.

### Plugin and Packer version

`v1.7.3`

This is happening in a GitHub Action running Ubuntu 20.04.3 LTS.

wedge-jarrad:
Allow Packer to create ed25519 keys as the temporary SSH keypair by setting the `temporary_key_pair_type` option. Defaults to rsa.

This is my first foray into the Packer code so I have no idea what I'm doing :D. Feel free to boss me around a bit if there are better ways to go about this, particularly the way the input is validated/defaulted.

I only spent a little bit of time looking at the tests but it wasn't immediately obvious to me how to incorporate this in the tests. Might have some more time to look a bit harder tomorrow (pointers appreciated!). I did, however, test manually and it seems to work. If you specify `temporary_key_pair_type = "ed25519"` you get an ed25519 keypair. Any other value or omission of `temporary_key_pair_type` results in an rsa keypair.

Closes #144
Relates https://github.com/hashicorp/packer/issues/10074


builder/common/step_key_pair.go
```diff
@@ -64,6 +64,7 @@ func (s *StepKeyPair) Run(ctx context.Context, state multistep.StateBag) multist
 	ui.Say(fmt.Sprintf("Creating temporary keypair: %s", s.Comm.SSHTemporaryKeyPairName))
 	keypair := &ec2.CreateKeyPairInput{
 		KeyName: &s.Comm.SSHTemporaryKeyPairName,
+		KeyType: &s.Comm.SSHTemporaryKeyPairType,
```
JenGoldstrich, line 67: Yep that was the issue!

JenGoldstrich (commented):
Looks great so far, as you pointed out there isn't a great place to test this currently, 

@nywilken and I are going to make some changes to the key_pair step code to make it easier to test, the refactor to allow us to mock the ec2Connection seems a bit complicated for a first time contributor, we will update you here once we have merged those changes so you can write a test.

We will also need to update the docs here https://github.com/hashicorp/packer/blob/master/website/content/partials/packer-plugin-sdk/communicator/SSHTemporaryKeyPair-not-required.mdx but we can handle that ourselves once we have your PR merged.

Thank you so much for this contribution, we'll get back to you soon!

wedge-jarrad:
Wasn't really sure what to do with the docs. The option is [already documented](https://www.packer.io/plugins/builders/amazon/ebs#temporary_key_pair_type) even though it hadn't been implemented and has fewer actual options than shown. The docs are generic so there is stuff there that doesn't apply to AWS. `temporary_key_pair_bits`, for example, isn't something that can be implemented for AWS since they don't give you that option when creating key pairs. Can't just remove those things, though, because they might be relevant and implemented for other plugins.

Thanks for working on the unit tests. Will keep an eye out for that when it's ready :)

JenGoldstrich:
Hey @wedge-jarrad,

I just merged https://github.com/hashicorp/packer-plugin-amazon/pull/181 which @nywilken and I worked on to add a unit test and an acceptance test for the keypair step.  So you'll need to rebase your PR on top of that.

For unit tests you should be able to un-comment the extra lines in this test https://github.com/hashicorp/packer-plugin-amazon/blob/06fe648380f11b12d2676bf4d20df67df7a413f2/builder/common/step_key_pair_test.go#L55 to test that the step key pair actually passes in the type as we'll expect it to, I would also add a unit test to this file as well https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/common/run_config_test.go to cover the default logic to rsa.

For acceptance tests if you'd like you can add a duplicate of this test https://github.com/hashicorp/packer-plugin-amazon/blob/06fe648380f11b12d2676bf4d20df67df7a413f2/builder/ebs/builder_acc_test.go#L495 to check if that if you pass in ed25519, it correctly sets that as the ssh-key type.  To run all acceptance tests you can run `make testacc`, or to run a single test, like the RSA one, you could run `PACKER_ACC=1 go test -count=1 -v ./... -run='TestAccBuilder_EbsKeyPair' --timeout=120m`

One other small change I'd recommend is to edit the run_config to include a check that the SSH Key type is rsa, or ed25519, and throw an error if its not one of those types, and then add a corresponding unit test.  Here is an example of a validator in the run config that checks the value of a field https://github.com/hashicorp/packer-plugin-amazon/blob/06fe648380f11b12d2676bf4d20df67df7a413f2/builder/common/run_config.go#L631

We can handle the docs change if you'd like.

I know this is a bit long so, feel free to let us know if there's any of that you'd like help on/are confused on!

wedge-jarrad:
Ok, I think that's everything. Let me know if I've missed anything :)

Acceptance tests:

```
$ PACKER_ACC=1 go test -count=1 -v ./builder/ebs/... -run='TestAccBuilder_EbsKeyPair' --timeout=120m
=== RUN   TestAccBuilder_EbsKeyPair_rsa
--- PASS: TestAccBuilder_EbsKeyPair_rsa (108.29s)
=== RUN   TestAccBuilder_EbsKeyPair_ed25519
--- PASS: TestAccBuilder_EbsKeyPair_ed25519 (103.57s)
PASS
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebs	211.876s
?   	github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance	[no test files]
```

Unit tests:

```
$ gmake test
?   	github.com/hashicorp/packer-plugin-amazon	[no test files]
ok  	github.com/hashicorp/packer-plugin-amazon/builder/chroot	0.224s
ok  	github.com/hashicorp/packer-plugin-amazon/builder/common	0.099s
?   	github.com/hashicorp/packer-plugin-amazon/builder/common/awserrors	[no test files]
?   	github.com/hashicorp/packer-plugin-amazon/builder/common/ssm	[no test files]
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebs	0.263s
?   	github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance	[no test files]
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebssurrogate	0.086s
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebsvolume	0.091s
ok  	github.com/hashicorp/packer-plugin-amazon/builder/instance	0.143s
ok  	github.com/hashicorp/packer-plugin-amazon/datasource/ami	0.054s
ok  	github.com/hashicorp/packer-plugin-amazon/datasource/parameterstore	0.047s
ok  	github.com/hashicorp/packer-plugin-amazon/datasource/secretsmanager	0.047s
?   	github.com/hashicorp/packer-plugin-amazon/post-processor/import	[no test files]
?   	github.com/hashicorp/packer-plugin-amazon/version	[no test files]
```

azr:
Super nice !

jsf9k:
Thanks for taking care of this @wedge-jarrad!

JRemitz:
Not sure if I'm doing something wrong, but I now see that my local private key is `ed25519` however packer fails to connect. If I connect with instance-connect in a debug session, I see the public key is still listed as `ssh-rsa` instead of `ssh-ed25519`.  

```sh
ssh-keygen -l
...
256 SHA256:....  (ED25519)
```

**Default ssh user's authorized_keys**
```sh
ssh-rsa AAAAB3NzaC1yc2... packer_62....
```

Feels like there is a mismatch here?

JRemitz:
Okay, to the point of this PR - it works great, thank you!  I was finally able to verify.

Building on my comment above, I'm not sure where to go but I'm building an AMI on top of another packer-built AMI.  It turns out that the public ssh key from the first Packer run is still in the `authorized_keys` and for whatever reason, the second Packer key isn't being added.  So my first key that was an `ssh-rsa` from a community image since Packer works with `ed25519` but Ansible provisioner fails.  My second Packer build on top of that fails with Packer unable to ssh to the temporary ec2 instance because the only Packer key that is being allowed/put into the `authorized_keys` on the machine was the public ssh key from the first build, despite the new keypair from AWS being the temporary key.  I was finally able to get this working by adding `ssh_clear_authorized_keys=true` so the key was cleared out after the first build. 😕

sbraz:
Hi, would it please be possible to fix this for the OpenStack builder as well? See https://github.com/hashicorp/packer-plugin-openstack/issues/54.

PR #180, opened 2022-01-20
Docs update: add note for block_duration_minutes

### LINKED ISSUE #175: Optional attributes for "object" types in variables?
#### Community Note

#### Description

I'd like to use "object" types for variables, but where some of the attributes are not required. I'd expect, when not specified, they would evaluate to the `default` value of the attribute. As a further example of the idea, Terraform has a similar concept, that they're [currently running as an experiment](https://www.terraform.io/language/expressions/type-constraints#experimental-optional-object-type-attributes).

#### Potential configuration

```
variable "example" {
  type = object({
    a = string           # a required attribute
    b = optional(string) # an optional attribute
  })
  default = {
    a = "foo"
    b = "bar"
  }
}
```

### LINKED ISSUE #176: amazon-ebs - "block_duration_minutes" not available for new AWS accounts
#### Overview of the Issue

I have been using Packer with the `block_duration_minutes` configuration for EBS builder for a while now. We use this feature to ensure that EC2 builders are terminated even when packer can't clean up for whatever reason.

Today I found an issue on a new AWS account and couldn't build an AMI. The error is:

```sh
==> amazon-ebs:
   Error waiting for fleet request (fleet-xxxx-xxxx-xxxx-xxxx-xxxx) to become ready:BlockDurationMinutes is not a valid parameter.
   BlockDurationMinutes is not a valid parameter.BlockDurationMinutes is not a valid parameter.
   No default subnet for availability zone: 'xxxx'.
   BlockDurationMinutes is not a valid parameter.BlockDurationMinutes is not a valid parameter.
```
* Please disregard the part about the default subnet.

After searching for a while I found this on AWS docs ([Define a duration for your Spot Instances](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances))

> Spot Instances with a defined duration (also known as Spot blocks) are no longer available to new customers from July 1, 2021. For customers who have previously used the feature, we will continue to support Spot Instances with a defined duration until December 31, 2022.

My suggestion for now would be to update the documentation for the parameter ([ebs#block_duration_minutes](https://www.packer.io/plugins/builders/amazon/ebs#block_duration_minutes)) so everyone is aware of this.

#### Reproduction Steps

On a new AWS account, or in an existing one that hasn't used spot instances with the `block_duration_minutes` configuration, try to build any AMI with the `amazon-ebs` builder and the parameter `block_duration_minutes` configured.

### Packer version

Packer v1.7.8

GMartinez-Sisti:
Hi, this updates the documentation for the parameter (ebs#block_duration_minutes) so everyone is aware of the it's deprecation to new customers.

Please do suggest changes to the way it's phrased if you think it can be improved.

Closes #176

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -93,7 +93,9 @@ type RunConfig struct {
 	// required duration for the Spot Instances (also known as Spot blocks). This
 	// value must be a multiple of 60 (60, 120, 180, 240, 300, or 360). You can't
 	// specify an Availability Zone group or a launch group if you specify a
-	// duration.
+	// duration. Note: This parameter is no longer available to new customers
+	// from July 1, 2021. [See Amazon's
+	//documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/spot-requests.html#fixed-duration-spot-instances).
 	BlockDurationMinutes int64 `mapstructure:"block_duration_minutes" required:"false"`
 	// Packer normally stops the build instance after all provisioners have
 	// run. For Windows instances, it is sometimes desirable to [run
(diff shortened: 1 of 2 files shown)


azr (approved):
Good catch, thanks for adding this !

PR #181, opened 2022-01-21
Add Unit test for KeyPair step

JenGoldstrich:
After reviewing https://github.com/hashicorp/packer-plugin-amazon/pull/179 with Wilken we came to the conclusion that we should add a unit test to the KeyPair step so that the aforementioned PR can be tested (just need to remove comments in step_key_pair_test)

I ended up having to make a small change to each builder to put the region in the state before the steps are run, previously step_key_pair would access Region through the ec2 object (ec2conn.Config.Region), since we now cast that as an ec2iface to be able to mock it in the test, we are no longer able to access the Config object within the step, since the interface does not know about the Config object (might be an upstream fix there)

### DIFF
--- a/builder/common/step_key_pair_test.go
+++ b/builder/common/step_key_pair_test.go
@@ -19,7 +19,6 @@ type mockEC2KeyPairConn struct {
 
 	CreateKeyPairCount int
 	CreateKeyPairArgs  []ec2.CreateKeyPairInput
-	waitCount          int
 
 	lock sync.Mutex
 }
--- a/builder/common/step_key_pair_test.go
+++ b/builder/common/step_key_pair_test.go
@@ -7,7 +7,7 @@ import (
 	"sync"
 	"testing"
 
-    "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/hashicorp/packer-plugin-sdk/communicator"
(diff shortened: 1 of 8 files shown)


builder/common/step_key_pair_test.go
```diff
@@ -0,0 +1,84 @@
...
+		Writer: new(bytes.Buffer),
+	})
+	conn := getKeyPairMockConn()
+	state.Put("ec2", conn)
+	state.Put("region", "us-east-1")
+	return state
+}
+
+//Add "AndType" to the name of this test
+func TestStepKeyPair_callsCreateKeyPairOnceWithName(t *testing.T) {
```
nywilken, line 55: I would add an additional set of tests and rename this one a bit, as I believe you want to test that the correct inputs are being added to the CreateKeyPairInput type. 

```suggestion
func TestStepKeyPair_withDefault(t *testing.T) {
```
This should assert that the temp key name matches and that the default keytype is rsa

Then I would add an additional test maybe called `func TestStepKeyPair_withKeyType` that does the same as this test but checks that the key type is set to the expected value.

nywilken (changes requested):
I left a couple of suggestions but otherwise looks good.

nywilken (approved):
Happy we were able to talk abit about the changes. Approving, feel free to merge when ready.

JenGoldstrich:
Didn't realize that https://github.com/hashicorp/packer-plugin-amazon/pull/182 was targeted at this PR's branch (step-key-pair-test), merging this now to merge both PRs

I ran acceptance tests, and things hang on a test for me, but they hang on the same test on the main branch for me as well, I think this is due to my setup, the test we added works which I think would validate that we haven't broken anything with this change.

PR #182, opened 2022-01-21
Wilken/step key pair test

nywilken:
Just adding an acceptance test for the keypair. It is a bit of a tricky test because it doesn't actually describe the keypair via the api but it does validated against the authorized_keys. You can add a test case for the ed25519 case when the other PRs are merge.


JenGoldstrich (approved):
LGTM, reviewed this with Wilken and switched it to an RSA test which can pass currently, will encourage contributor of original PR to add acceptance test for ed25519

PR #183, opened 2022-01-24
Change error handling with CSV Parsing of filters

JenGoldstrich:
After https://github.com/hashicorp/packer-plugin-amazon/pull/174 @azr pointed out that we should bubble up errors from csv parsing, this PR bubbles up that error, and addresses some other docs feedback.

For testing I added checks that errors aren't thrown to both existing filtering unit tests, I felt like adding additional tests wasn't worth it for this error, we can decide to refactor the build filter class to accept a mocked csv parser if folks want to though.

Addresses feedback from: https://github.com/hashicorp/packer-plugin-amazon/pull/174#discussion_r790534015 and https://github.com/hashicorp/packer-plugin-amazon/pull/174#discussion_r790903152

### DIFF
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -140,7 +140,7 @@ type Config struct {
 	//This will fail unless *exactly* one AMI is returned. In the above example,
 	//`most_recent` will cause this to succeed by selecting the newest image.
 	//
-	//-   `filters` (map{key string, value string | comma separated string}) -
+	//-   `filters` (map[string,string] | multiple filters are allowed when seperated by commas) -
 	//  filters used to select a `source_ami`.
 	//	NOTE: This will fail unless *exactly* one AMI is returned. Any filter
 	//	described in the docs for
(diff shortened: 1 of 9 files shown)


JenGoldstrich:
Currently two acceptance tests are failing for me on acceptance testing, 

TestAccBuilder_EbsSessionManagerInterface
TestAccAmazonSecretsManager

Both of these tests fail for me when I go back to a previous commit of the plugin from a few weeks ago, so I suspect this is not related to recent changes, would appreciate some help validating that assumption though before releasing.

lorengordon:
Thank you so much for accepting our PR and then enhancing it @JenGoldstrich! Really makes the community contributor experience feel very welcoming and easy!

azr (approved):
Super nice, LGTM, thanks for the quick turnaround :D

azr:
We could see about the acc tests in another PR/issue

JenGoldstrich:
Merging based on Adrien's approval, will review acc tests separately prior to release.

PR #191, opened 2022-02-11
Remove default tagging

### LINKED ISSUE #155: Packer insist on putting "Packer Builder" tag
_This issue was originally opened by @grumpper in https://github.com/hashicorp/packer/issues/11370 and has been migrated to this repository. The original issue description is below._

---
#### Overview of the Issue

When I am running packer build it puts `Name = "Packer Builder"` on the temp instance used for building and on the AMI itself. This happens when I don't put `tags` and `run_tags` at all or even when explicitly set them to be `{}`.

#### Reproduction Steps

Define packer template and setup the `amazon-ebs` source.
Remove any tagging config from the source OR set it to `{}` explicitly.
Use the template to build something and observe how both the instance and the AMI created from it are tagged nevertheless.

### Packer version

Tested on packer `1.7.7` and `1.7.8`

### Simplified Packer Template
```
source "amazon-ebs" "stuff" {
  ami_name                      = "${var.ami_base_name}-${var.source_version}-hvm-${formatdate("YYYYMMDD", timestamp())}-x86_64-gp2"
/.../
  tags          = {}
  run_tags   = {}
/.../
}

build {
  sources = [
    "source.amazon-ebs.stuff.id"
  ]

  provisioner "shell" {
    inline = [
      "echo \"some script stuff\""
    ]
  }
}

macOS 12.0.1 (x86_64)

no errors in logs; it's just that this behavior is nonsense cause you have means to control the tags you want but you actually don't since packer decides to put one anyway no matter your config.
When building an AMI for clients who do not give permissions to tag AMIs since they want to control that themselves the build directly errors out. And there is no way to work around that.

azr:
This fixes #155 and a is breaking change.
We should probably bump the minor version for the next release.

### DIFF
--- a/builder/common/ami_config.go
+++ b/builder/common/ami_config.go
@@ -57,6 +57,8 @@ type AMIConfig struct {
 	// Key/value pair tags applied to the AMI. This is a [template
 	// engine](/docs/templates/legacy_json_templates/engine), see [Build template
 	// data](#build-template-data) for more information.
+	//
+	// The builder no longer adds a "Name": "Packer Builder" entry to the tags.
 	AMITags map[string]string `mapstructure:"tags" required:"false"`
 	// Same as [`tags`](#tags) but defined as a singular repeatable block
 	// containing a `key` and a `value` field. In HCL2 mode the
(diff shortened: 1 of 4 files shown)


nywilken (approved):
This looks good to me. Upon release we should update the release note to include a little more details around the removal of default Name tag. 

With versioned docs I do wonder if we should include a little call out for the tags field that a default is no longer being set. What do you think?

azr:
Good idea Wilken, adding that call out !

PR #195, opened 2022-02-17
docs: specify valid values for ami_groups

mweibel:
in https://github.com/hashicorp/packer/commit/b20e26be17b56e893d9b29945441041824d0e110 a limit for the ami_groups was added but that is not in this documentation anymore, even though the official docs suggest this is still the only value possible: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_LaunchPermission.html

### DIFF
--- a/builder/common/ami_config.go
+++ b/builder/common/ami_config.go
@@ -33,7 +33,8 @@ type AMIConfig struct {
 	AMIUsers []string `mapstructure:"ami_users" required:"false"`
 	// A list of groups that have access to
 	// launch the resulting AMI(s). By default no groups have permission to launch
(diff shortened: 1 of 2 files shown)


nywilken:
Hi @mweibel thanks for the fix. The documentation you updated is autogenerated by the code comments. But for some reason the checks did not run on the PR. So you've had no way of seeing that. I made the change and re-generated the docs with your fix. I'll get this merged in shortly.

nywilken (approved):
LGTM

mweibel:
ah! good to know. Thanks for updating & merging 👏

PR #199, opened 2022-03-05
Handle empty source tags in restricted regions

john-pierce:
Fix build failure when no source instance tags are present in restricted
regions.

e83d836e removed default tagging for source instances which introduces
the potential for launching an instance with an empty ec2Tags list.

This was handled for regions that support tagging on instance creation
but not in restricted regions (GovCloud and China) where tags are
applied to the instance after creation.

PR #200, opened 2022-03-05
Remove GovCloud from isRestricted run source param

john-pierce:
Tag-On Create for EC2 and EBS is now supported on GovCloud:
https://aws.amazon.com/about-aws/whats-new/2018/07/tag-on-create-for-amazon-ec2-and-amazon-ebs-is-now-available-in-the-aws-govcloud-region/

isRestrcted is only used to gate Tag-On Create and is no longer needed
for GovCloud.

### DIFF
--- a/builder/ebs/builder.go
+++ b/builder/ebs/builder.go
@@ -246,7 +246,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			InstanceMetadataTags:              b.config.Metadata.InstanceMetadataTags,
 			InstanceInitiatedShutdownBehavior: b.config.InstanceInitiatedShutdownBehavior,
 			InstanceType:                      b.config.InstanceType,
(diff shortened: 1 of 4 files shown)


john-pierce:
See #199 for build failure.

Log from #200 patch only:
```
==> amazon-ebs.no-tags: Launching a source AWS instance...
    amazon-ebs.no-tags: Instance ID: i-073f2d2b075517a4e
==> amazon-ebs.no-tags: Waiting for instance (i-073f2d2b075517a4e) to become ready...
2022/03/05 14:03:33 packer-plugin-amazon plugin: 2022/03/05 14:03:33 [INFO] Not using winrm communicator, skipping get password...
==> amazon-ebs.no-tags: Using SSH communicator to connect: 160.1.8.1
```

nywilken (approved):
LGTM - Thanks for the contribution

nywilken:
I noticed that the restriction was still being set for the ebssurrogate builder. I added the change to your branch before merging.

PR #208, opened 2022-03-29
Mention source config settings for AWS polling

ulfjack:
The default polling configuration is too short for AWS MacOS images due to AWS taking a very long time to shut down an instance and create an image. Instead of *always* setting the environment variables, it's much better to use the config file settings in that case (I don't know how often I forgot these env variables until I discovered that they could be set in the config file).


nywilken (approved):
@ulfjack thanks for the add. Great idea to have it mentioned right along side the ENVs.

Issue #226, opened 2022-06-03
Existing subnet_filter with quotes failing to parse after upgrade


jaredledvina:
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

After upgrading to Packer 1.8.1 (from 1.7.4), a `subnet_filter` that we've used for a while now is suddenly failing. 

#### Reproduction Steps

Configure a Packer AWS builder using Packer 1.8.1 with a `subnet_filter` like:

```
  subnet_filter {
    filters = {
      "tag:test" = "{\"purpose\":\"testing\"}"
    }
    random = true
  }
```

Packer will fail with:
```
Build 'amazon-ebs.ebs' errored after 1 second 173 milliseconds: Couldn't parse subnet filters: parse error on line 1, column 3: extraneous or missing " in quoted-field
```

### Plugin and Packer version

Packer v1.8.1

### Simplified Packer Buildfile

I can work on this, currently it's our prod file which I can't share directly. 

### Operating system and Environment details

Linux x86_64 Gitlab runner

### Log Fragments and crash.log files

```
amazon-ebs.ebs: output will be in this color.
==> amazon-ebs.ebs: Prevalidating any provided VPC information
==> amazon-ebs.ebs: Prevalidating AMI Name: snip
    amazon-ebs.ebs: Found Image ID: ami-snip
Build 'amazon-ebs.ebs' errored after 1 second 208 milliseconds: Couldn't parse subnet filters: parse error on line 1, column 2: bare " in non-quoted-field
==> Wait completed after 1 second 208 milliseconds
==> Some builds didn't complete successfully and had errors:
--> amazon-ebs.ebs: Couldn't parse subnet filters: parse error on line 1, column 2: bare " in non-quoted-field
==> Builds finished but no artifacts were created.
```

PR #227, opened 2022-06-03
Enable csv LazyQuotes in build_filter

jaredledvina:
Currently, if a filter has quotes in the value, it fails with:

```
Build 'amazon-ebs.ebs' errored after 1 second 208 milliseconds: Couldn't parse subnet filters: parse error on line 1, column 2: bare " in non-quoted-field
```

I believe this is because https://github.com/hashicorp/packer-plugin-amazon/pull/174. This adds a test to confirm and then enables `LazyQuotes` which I got the idea from https://stackoverflow.com/questions/31326659/golang-csv-error-bare-in-non-quoted-field. The test now passes.  

Closes https://github.com/hashicorp/packer-plugin-amazon/issues/226


nywilken (approved):
Nice catch. The changes look good to me.

PR #243, opened 2022-07-20
Change AssociatePublicIpAddress to a bool pointer

### LINKED ISSUE #240: Unable to set AssociatePublicIpAddress to false
#### Overview of the Issue

We use a org-wide SCP rule blocking creation of EC2 with public IP. When setting `associate_public_ip_address` to false, this if statement is never taken:

https://github.com/hashicorp/packer-plugin-amazon/blob/c73a909478039431f94c06e2a0e27bb84ec9a98d/builder/common/step_run_source_instance.go#L186

By default AWS API sets `AssociatePublicIpAddress: true`, and later checks what is configured on the subnet. (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_InstanceNetworkInterfaceSpecification.html) 

`AssociatePublicIpAddress` variable should be a pointer, making it a tristate variable. Then I can explicit set `AssociatePublicIpAddress: false`, and my SCP rule will not deny the action. 

#### Reproduction Steps

Create a SCP rule on AWS:
```    
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Deny",
      "Action": [
        "ec2:RunInstances"
      ],
      "Condition": {
        "BoolIfExists": {
          "ec2:AssociatePublicIpAddress": "true"
        }
      },
      "Resource": "arn:aws:ec2:*:*:network-interface/*"
    }
  ]
}
```

Try to run packer.

### Plugin and Packer version

Any packer version

Any Buildfile with `associate_public_ip_address = false`.

pedrokiefer:
Change AssociatePublicIpAddress to a bool pointer, making it a tristate variable as it's already defined on HCL. 

This enables to explicitly set a `NetworkInterface` with `AssociatePublicIpAddress` set to false.

We tested the built plugin on our AWS account with the SCP rule mentioned on #240, it works perfectly. 

Closes #240

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ main
 dist/*
 packer-plugin-amazon
 .docs
+.DS_Store
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/hashicorp/packer-plugin-sdk/communicator"
 	"github.com/hashicorp/packer-plugin-sdk/template/config"
+	confighelper "github.com/hashicorp/packer-plugin-sdk/template/config"
 	"github.com/hashicorp/packer-plugin-sdk/template/interpolate"
 	"github.com/hashicorp/packer-plugin-sdk/uuid"
 )
@@ -88,8 +89,8 @@ type MetadataOptions struct {
 type RunConfig struct {
 	// If using a non-default VPC,
 	// public IP addresses are not provided by default. If this is true, your
-	// new instance will get a Public IP. default: false
-	AssociatePublicIpAddress bool `mapstructure:"associate_public_ip_address" required:"false"`
+	// new instance will get a Public IP. default: unset
+	AssociatePublicIpAddress confighelper.Trilean `mapstructure:"associate_public_ip_address" required:"false"`
 	// Destination availability zone to launch
 	// instance in. Leave this empty to allow Amazon to auto-assign.
 	AvailabilityZone string `mapstructure:"availability_zone" required:"false"`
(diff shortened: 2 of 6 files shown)


nywilken:
Hi @pedrokiefer the Packer SDK has a type just for these cases. It's the Trilean type. I created a branch using the SDK type to compare your changes against https://github.com/hashicorp/packer-plugin-amazon/commit/67be6e0b4659b6de19b4ae7a8cc617c92527d2d6. 

Does using this type resolve the issue you are running into? 

That said, this changes is also related https://github.com/hashicorp/packer-plugin-amazon/issues/18 could you validate if the change also fixes the issues being reported.

pedrokiefer:
Hi @nywilken! Sorry for the late reply, took some days off. 

I didn't know about the Trilean type (nice to have that!), so I used the old and good `*bool`. Your changes to Trilean do fix my issue with packer.

Reading about #18, it seems to be a different issue, ie. setting `AssociatePublicIpAddress` without a `SubnetId` — so it probably needs its own patch / logic fix.

nywilken:
> I didn't know about the Trilean type (nice to have that!), so I used the old and good *bool. Your changes to Trilean do fix my issue with packer.

@pedrokiefer thanks for validating I'll go ahead and take care of merging my changes with your PR so that you get the change credit. Thanks for making the fix. 

As for the Trilean it was more of thing when JSON was the only supported template type with HCL2 that changes a bit so we don't see a lot of its use.

nywilken (approved):
Pardon the force push @pedrokiefer. I updated your changes to use the Trilean type and rebased to get in the latest circle configuration. I'll merge once all goes green.

PR #244, opened 2022-07-26
Remove tags from InstanceMetadataOptionsRequest when in the default state to fix launch issue for unsupported regions

### LINKED ISSUE #242: Instance metadata tags are not supported in all regions
#### Overview of the Issue

Since the merge of this PR https://github.com/hashicorp/packer-plugin-amazon/pull/194 packer no longer works in regions not supporting instance metadata tags. A region like this is `us-iso-east-1`. 
The error from AWS is `Setting InstanceMetadataTags is not yet supported in this region`.

#### Reproduction Steps

Run packer in a region not supporting instance metadata tags.

### Plugin and Packer version

From `packer version` 1.8.1 and AWS plugin 1.0.9

nywilken:
The ability to include instance tags within the instance metadata
service is not supported for non-commercial regions such as GovCloud.
This change will only include the configuration for instances metadata
tags when enabled. By default, where supported, unspecified settings
will use the default API value of "disabled". Enabling support for tags
in a unsupported region will result in an AWS launch error, which will
terminate the build.

Closes #242

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -78,6 +78,7 @@ type MetadataOptions struct {
 	// Defaults to 1.
 	HttpPutResponseHopLimit int64 `mapstructure:"http_put_response_hop_limit" required:"false"`
 	// A string to enable or disable access to instance tags from the instance metadata. Defaults to disabled.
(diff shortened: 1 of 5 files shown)


sylviamoss (approved):
LGTM 👍🏼

PR #250, opened 2022-08-01
Allow restricting temporary SG to current host's public IP

### LINKED ISSUE #46: Restrict AWS Security group to the current host's Public IP address according to AWS
_This issue was originally opened by @tedivm as hashicorp/packer#9915. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Community Note

#### Description

AWS provides an extremely simple service that you can use to find out what IP address AWS from your machine.

https://checkip.amazonaws.com/

#### Use Case(s)

This would make it so the machine did not have to be open to the world.

#### Potential References

#6735

r-bk:
Add support for restricting temporary SG to current host's public IP instead of `0.0.0.0/0`.

### Motivation

Currently packer creates a temporary security group for an EC2 instance. And, by default, it opens the SSH port to any source IPv4 address (i.e., `0.0.0.0/0`). Such behavior is discouraged in many communities and organizations.

### Proposed Solution

This PR adds a new configuration option `temporary_security_group_source_public_ip`, which supplements the already existing option `temporary_security_group_source_cidrs`. By default, when `temporary_security_group_source_cidrs` is not specified, packer uses `0.0.0.0/0` to allow access to the EC2 instance. Now, if `temporary_security_group_source_public_ip` is set to `true`, packer will try to determine current host's public IP by issuing a GET request to https://checkip.amazonaws.com, and use it to allow access to the EC2 instance, instead of opening the instance to the world (`0.0.0.0/0`).

This option achieves the motivation behind this PR, because, when enabled, the temporary SG is restricted to a specific IP of the packer's host, and doesn't open the SSH port to the world.

Closes #46


r-bk:
@tedivm @nywilken 

This PR handles the issue https://github.com/hashicorp/packer/issues/9915 in which you have participated.
Your review would be greatly appreciated.

nywilken:
@r-bk thanks for opening up this change we will take a look this week and get back to you with any feedback and suggestions (if any). Cheers!

builder/common/step_security_group.go
```diff
@@ -136,10 +137,39 @@ func (s *StepSecurityGroup) Run(ctx context.Context, state multistep.StateBag) m
 
 	log.Printf("[DEBUG] Found security group %s", s.createdGroupId)
 
+	temporarySGSourceCidrs := s.TemporarySGSourceCidrs
+	if len(temporarySGSourceCidrs) == 0 && s.TemporarySGSourcePublicIp {
```
lbajolet-hashicorp, line 141: I'd suggest warning the user that if both arguments are used together, one will be ignored. This will prevent a source of confusion for users.

```suggestion
	if len(temporarySGSourceCidrs) != 0 && s.TemporarySGSourcePublicIp {
		ui.Warn("Using temporary_security_group_source_public_ip with temporary_security_group_source_cidrs is unsupported, ignoring temporary_security_group_source_public_ip.")
	}
	if len(temporarySGSourceCidrs) == 0 && s.TemporarySGSourcePublicIp {
```

I may be overcareful though, the documentation is pretty clear on what is expected, still I think this will help prevent surprises if we clearly state what the problem with the configuration is.

builder/common/step_security_group.go
```diff
@@ -136,10 +137,39 @@ func (s *StepSecurityGroup) Run(ctx context.Context, state multistep.StateBag) m
...
+			err := fmt.Errorf("Current host's public IP is unspecified: %s", ip)
+			ui.Error(err.Error())
+			state.Put("error", err)
+			return multistep.ActionHalt
+		}
+
+		ui.Say(fmt.Sprintf("Current host's public IP: %s", ip))
+
+		bits := 128
+		if tmp := ip.To4(); tmp != nil {
```
lbajolet-hashicorp, line 162: This is somewhat misleading here, and to be clear this is not your fault, it's mostly on `ip.To4()` that attempts to parse the ip as V6 (i.e. 4 bytes) instead of v4, but may I suggest adding a little comment to clarify this bit of code?

```suggestion
		// ip.To4 attempts to parse the IP as IPv6, if this fails, we fallback to IPv4.
		if tmp := ip.To4(); tmp != nil {
```

lbajolet-hashicorp (approved):
Hey @r-bk,

The code looks good to me, I just have a couple suggestions, but they're essentially nitpicks more than requests for changes, feel free to respond and either reject or accept my suggestions.

Other than that, I'd say we can merge this soon.

Sorry for the delay in reviewing your code, thanks for this feature, much appreciated.

builder/common/step_security_group.go
r-bk: At this point in code I do not know if the IP address I received from AWS is v4 or v6.
All I know is that it is one of them.
 To format a correct CIDR I must choose the correct bitmask according to the address type.
`ip.To4()` returns a value if the address is v4, or `nil` otherwise (which means it is v6). So this is my test for choosing the correct bitmask.

> `// ip.To4 attempts to parse the IP as IPv6, if this fails, we fallback to IPv4.`

I am not sure I understand this comment, cause `To4` checks if the address is v4, it doesn't try to parse it as v6.

Did you mean `// ip.To4() attempts to parse the IP as IPv4, if this fails, we fallback to IPv6.` ?

builder/common/step_security_group.go
r-bk: Sure. Will add.

r-bk: done

builder/common/step_security_group.go
r-bk: @lbajolet-hashicorp I have added a comment which I think clarifies the code.
Please tell me if that is OK.

lbajolet-hashicorp: Oh you're right, I misunderstood the function as attempting to convert to IPv6, but it is indeed IPv4, sorry for that.
The corrected comment looks good to me

lbajolet-hashicorp:
All good to me, we're merging this, thanks again for the contribution @r-bk !

PR #257, opened 2022-08-19
Add `enable_nitro_enclave` argument to enable support for Nitro Enclaves on instance

### LINKED ISSUE #245: Add support for EC2 Nitro Enclaves
_This issue was originally opened by @urluba in https://github.com/hashicorp/packer/issues/11895 and has been migrated to this repository. The original issue description is below._

---

#### Community Note

* Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request
* Please do not leave "+1" or other comments that do not add relevant new information or questions, they generate extra noise for issue followers and do not help prioritize the request
* If you are interested in working on this issue or have submitted a pull request, please leave a comment

Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

#### Description

Nitro Enclave creates an isolated environment on the EC2 instance. This environment has its own kernel, processor and memory and can only communicate with the parent instance through a local socket.

Support of Nitro enclave is disabled by default and requires to be enabled at launch or on a stopped instance.
Packer currently does not allow to create an EC2 Instance with Nitro Enclave enabled.

#### Use Case(s)

Having an EC2 Instance with Nitro Enclave enabled allows us to use certificates saved in ACM with Nginx preventing the need to store sensitive date in the AMI. Thus someone able to retrieve the AMI don't leave with the private keys and/or passphrase.

#### Potential configuration

```
source "amazon-ebs" "example" {
    enclave_support = True
}
```

SajanAlexander-okta:
Adds support for enabling Nitro Enclaves on instance.  An example config would be: 

```
source "amazon-ebs" "example" {
    enable_nitro_enclave = true
}
```

Closes #245

### DIFF
--- a/builder/instance/builder.go
+++ b/builder/instance/builder.go
@@ -211,11 +211,12 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 			errs, fmt.Errorf("x509_key_path points to bad file: %s", err))
 	}
 
-	if b.config.IsSpotInstance() && ((b.config.AMIENASupport.True()) || b.config.AMISriovNetSupport) {
+	if b.config.IsSpotInstance() && ((b.config.AMIENASupport.True()) || b.config.AMISriovNetSupport || b.config.EnableNitroEnclave) {
 		errs = packersdk.MultiErrorAppend(errs,
 			fmt.Errorf("Spot instances do not support modification, which is required "+
-				"when either `ena_support` or `sriov_support` are set. Please ensure "+
-				"you use an AMI that already has either SR-IOV or ENA enabled."))
+				"when either `ena_support`, `sriov_support`, or `enable_nitro_enclave` "+
+				"are set. Please ensure you use an AMI that already has either SR-IOV "+
+				"or ENA enabled."))
 	}
 
 	if b.config.RunConfig.SpotPriceAutoProduct != "" {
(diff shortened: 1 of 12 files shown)


builder/instance/builder.go
```diff
@@ -298,6 +299,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			Ctx:                      b.config.ctx,
 			Debug:                    b.config.PackerDebug,
 			EbsOptimized:             b.config.EbsOptimized,
+			EnableNitroEnclave:       b.config.EnableNitroEnclave,
```
nywilken, line 302: Is there a specific reason why this was only added to the amazon-instance builder? 

It looks like this feature is only enabled for the instance builder. This same structure change to `awscommon.StepRunSourceInstance` needs to be added to the Ebs type builders in order to support Enclave for all builder types.

nywilken (changes requested):
Hi @SajanAlexander-okta thanks for opening up this change. This is off to a good start. Support for Enclave should be added to all the other supported builders. If for some reason it can not be added for the ebssurrogate or ebsvolume builders it should be called out in the documentation.

builder/instance/builder.go
SajanAlexander-okta: That was just an oversight.  I had got it working with the instance builder and forgot to go back to the other ones.

nywilken (approved):
Hi @SajanAlexander-okta thanks for the update here. This looks good to me. I'll merge, and should have it released later this week barring any open blockers.

SajanAlexander-okta:
Thank you, @nywilken!

PR #260, opened 2022-08-23
Create fix for deprecation of images in other regions

wampiedriessen:
These changes should fix the crash that occurs when enabling deprecation while also copying images to other regions.

Closes #259 

This PR is still a work-in-progress

PR #262, opened 2022-08-31
Depreciate copied image fix

### LINKED ISSUE #259: Enable AMI deprecation: InvalidAMIID.NotFound: Invalid ImageId
#### Overview of the Issue

When copying a created AMI to multiple regions with ```deprecate_at``` set, packer fails with:
```
eks-packer-ami-build.amazon-ebs.this: Error enable AMI deprecation: InvalidAMIID.NotFound: Invalid ImageId: ami-0123456789
==> eks-packer-ami-build.amazon-ebs.this: 	status code: 400, request id: xxxxx
```

This looks related to:
##### Previous:
* https://github.com/hashicorp/packer-plugin-amazon/issues/137

##### Recent:
* https://github.com/hashicorp/packer-plugin-amazon/issues/249
* https://github.com/hashicorp/packer-plugin-amazon/pull/251

Issue is present in plugin v1.1.3, and not in v1.1.2

The difference between the two is that in 1.1.2, Deprecation is added to the image _before_ copying, and in 1.1.3, Deprecation is added _after_ the copy.

#### Reproduction Steps

Build an AMI, with ```ami_regions``` as a list of more than one region, and ```deprecate_at``` set.

### Plugin and Packer version

* Packer 1.8.3 (also tested with 1.7.9)
* Plugin version 1.1.3

```
locals {
  expiration_date_hours = var.ami_expiration_days * 24
  expiration_date = timeadd(timestamp(), "${local.expiration_date_hours}h")
}

source "amazon-ebs" "this" {
  ami_name = "myami"
  ami_description = "My AMI"
  ami_users = var.aws_account_ids
  ami_regions = var.aws_regions
  ami_virtualization_type = "hvm"
  ebs_optimized = true
  encrypt_boot = false
  deprecate_at = local.expiration_date
  associate_public_ip_address = false
  vpc_filter {
    filters = {
      "tag:Name": var.vpc_tag_name,
      isDefault: false
    }
  }
  subnet_filter {
    filters = {
      "tag:Name": var.subnet_tag_name
    }
  }
  security_group_filter {
    filters = {
      "tag:Name": var.security_group_name
    }
  }
  source_ami_filter {
    filters = {
        <INSERT_FILTERS_HERE>
    }
    most_recent = true
  }
  instance_type = "m5.xlarge"
  ssh_username = "ec2-user"
}

build {
  name = "eks-packer-ami-build"
  sources = ["source.amazon-ebs.this"]
}
```

Linux x86_64

```
TBD
```

### Workaround

Pinning the version to ```1.1.2``` resolves this for now:

```
packer {
  required_plugins {
    amazon = {
      version = "= 1.1.2"
    }
  }
}
```

lbajolet-hashicorp:
A template with deprecation on a copied image would fail as-is the depreciation API call on AWS does not accept a region as parameter.

A previous attempt to fix this was merged in PR #138, whence the depreciation step was moved before copying, as copying an AMI from one region to another copies the depreciation information along with it.

This failed however with encrypted AMIs as encrypting the image after depreciation failed to propagate the attribute to the encrypted image.

Hence, this commit is an attempt to fix both situations, the step remains after copying/encrypting the image, and connects to each region in order to depreciate every copy made of the original AMI.

Closes #259


builder/ebs/builder_acc_test.go
```diff
@@ -76,6 +76,73 @@ func TestAccBuilder_EbsRegionCopy(t *testing.T) {
 	acctest.TestPlugin(t, testCase)
 }
 
+func TestAccBuilder_EbsRegionsCopyWithDeprecation(t *testing.T) {
```
nywilken, line 79: > This failed however with encrypted AMIs as encrypting the image after depreciation failed to propagate the attribute to the encrypted image.

Given this statement would it makes sense to add an acceptance test with encryption enabled to ensure that the copy and deprecation attribute are set as expected?

builder/ebs/builder_acc_test.go
```diff
@@ -76,6 +76,73 @@ func TestAccBuilder_EbsRegionCopy(t *testing.T) {
...
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+
+			err := checkRegionCopy(amiName, []string{"us-east-1", "us-west-1"})
+			if err != nil {
+				return err
+			}
+
+			var errors []error
```
nywilken, line 121: This is not a suggested changed I'm just sharing for your information we have a pkg for capturing multi-errors github.com/hashicorp/go-multierror that we use within Packer plugins for things like.

nywilken (commented):
This change looks great. I have not pulled do to run acceptance test yet. But the changes look good. Nice addition on the test case. 

I left a question given the encryption statement you have in your description but this looks otherwise good.

builder/ebs/builder_acc_test.go
lbajolet-hashicorp: The previous PR from @wampiedriessen added such a test to the suite, I can consider adding a new one that ensures the combination of copy + encrypt + depreciate does work, I'll add this to the PR before we can merge it

builder/ebs/builder_acc_test.go
lbajolet-hashicorp: Great suggestion, I had no idea this existed, I'll adapt my code to use it, thanks!

builder/ebs/builder_acc_test.go
lbajolet-hashicorp: Test for encryption + copy + depreciate added in the latest reroll

wampiedriessen:
Hey @lbajolet-hashicorp This PR is looking good!

I have no comments regarding your changes, LGTM.

During my own attempt at this fix (#260) I was wondering whether or not we should move `step_enable_deprecation` into the `common` package altogether.
As far as I understand this step could also be applied to the other AMI builders this packer repository has to offer.

Maybe that refactor might be a little out of scope, but if you agree we could write up an issue.

Cheers!
Wampie

lbajolet-hashicorp:
Hi @wampiedriessen,

Good observation, I think it may make some sense indeed to move it to common if some other builder wants to leverage it. For the moment only EBS implements that, but moving the step to common will make it possible for `GetRegionConn` to become private again.

It can be done in a later PR indeed, opening an issue to keep that in mind would be appropriate I think. We can work on opening one after this PR is merged.

Thanks for the suggestion!

dgibbard-cisco:
Tested and working for me 👍

nywilken (approved):
The updated test case looks good. Nicely done. Feel free to merge when ready.

dgibbard-cisco (approved):

PR #263, opened 2022-08-31
Fix issues reported by Go checks

nywilken:


builder/chroot/step_attach_volume.go
```diff
@@ -14,10 +14,6 @@ import (
 
 // StepAttachVolume attaches the previously created volume to an
 // available device location.
-//
-// Produces:
-//   device string - The location where the volume was attached.
-//   attach_cleanup CleanupFunc
```
lbajolet-hashicorp, line 20: I'm surprised this needs to be removed, I think this comment brings some relevant information about the steps, it's sad to see it go. Does the linter complain about this? If so, is it something we can work around?

lbajolet-hashicorp (approved):
LGTM, just have a little comment on the removal of the `Produces` sections in the comments for steps, which may also be applicable to the removal of the config examples for `AssumeRoleConfig`. I think it's sad to remove them, especially if the linter complains, they bring valuable information imho, so unless they're deemed useless, I'd prefer to keep them.

Other than that, :shipit: !

builder/chroot/step_attach_volume.go
nywilken: The latest version of the linter does complain. I suspect that Go 1.19 will have something to say about these comments, as they containing multiples spaces after the double slashes. Even with the multi-line comment operator the linter throws a warning. I fixed this by downgrading the linter for now. 

It will give us time to see what can be done other than skipping them by putting `//nolint:goimports` for these types of doc comments. I'm partly thinking this might be a bug with the linter but I haven't had a chance to dive in.

lbajolet-hashicorp: Hm, if the problem is with the multiple spaces, would the linter accept starting the line with ` * ` instead? I figure it'd fix the issue for now so we won't have to remove those comments

lbajolet-hashicorp (approved):
With the downgraded linter, the changes look good to me, we can merge when you want @nywilken!

PR #273, opened 2022-09-23
Add support for toggling unlimited credits for T3/T3a/T4g instance type

### LINKED ISSUE #127: Amazon Builders on t3 instances have unlimited credits enabled by default
_This issue was originally opened by @ricardclau in https://github.com/hashicorp/packer/issues/6721 and has been migrated to this repository. The original issue description is below._

---
This is related to https://github.com/hashicorp/packer/issues/6260 and it happens with all packer versions

AWS recently launched t3 instances which have better network performance (and are slightly cheaper) than the previous t2 family

However, they come with a hidden surprise, unlimited credits seems to be on by default

I guess it should be a relatively easy fix. Code in the mentioned issue should be slightly modified so that unlimited credits is explicitly set to false by default and only set to true if the packer user decides to

If nobody wants to pick this up, I may have some time to work on it!

nywilken:
- Add support to toggle unlimited credits for t3 and t4g instance types.
- Add deprecation warning for EnableT2Unlimited to all builders
- Add test validating EnableUnlimitedCredits is true when EnableT2Unlimited is true
- Update default credit type to standard for all burstable instance types
- Add EnableUnlimitedCredits for spot instance type builds
- Update StepRunSourceInstance to use CPUCredits const values
- Add acceptance test for enable_unlimited_credits

Closes #127

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -18,6 +18,12 @@ import (
 	"github.com/hashicorp/packer-plugin-sdk/uuid"
 )
 
+const (
+	// https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreditSpecificationRequest.html#API_CreditSpecificationRequest_Contents
+	CPUCreditsStandard  = "standard"
+	CPUCreditsUnlimited = "unlimited"
+)
+
 var reShutdownBehavior = regexp.MustCompile("^(stop|terminate)$")
 
 type SubnetFilterOptions struct {
@@ -821,13 +827,14 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 	}
 
 	if c.EnableUnlimitedCredits {
-		if c.SpotPrice != "" {
-			errs = append(errs, fmt.Errorf("Error: Unlimited credits cannot be used in conjunction with Spot Instances"))
-		}
-
 		if !c.IsBurstableInstanceType() {
 			errs = append(errs, fmt.Errorf("Error: Instance Type: %s is not within the supported types for Unlimited credits. Supported instance types are T2, T3, and T4g", c.InstanceType))
 		}
+
+		if c.SpotPrice != "" && regexp.MustCompile(`^t2\.`).MatchString(c.InstanceType) {
+			errs = append(errs, fmt.Errorf("Error: Unlimited credits cannot be used in conjunction with Spot Instances"))
+		}
+
 	}
 
 	var tenancy string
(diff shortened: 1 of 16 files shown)


builder/common/run_config.go
```diff
@@ -847,3 +877,10 @@ func (c *RunConfig) SSMAgentEnabled() bool {
 	hasIamInstanceProfile := c.IamInstanceProfile != "" || c.TemporaryIamInstanceProfilePolicyDocument != nil
 	return c.SSHInterface == "session_manager" && hasIamInstanceProfile
 }
+
+// IsBurstableInstanceType checks if the InstanceType for the config is one
+// of the following types T2, T3a, T3, T4g
+func (c *RunConfig) IsBurstableInstanceType() bool {
```
nywilken, line 883: The Terraform provider uses the DesribeInstanceOffering API here to determine if the selected instance is burstable. This is a nice approach for sure. It does introduce a new API call, which I'm not sure if it would introduce a regression as a new set of permissions may be needed. I can follow up on this.

builder/ebs/builder_acc_test.go
```diff
@@ -770,6 +770,40 @@ func TestAccBuilder_EbsKeyPair_rsaSHA2OnlyServer(t *testing.T) {
...
+			if buildCommand.ProcessState.ExitCode() != 0 {
+				return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+			}
+			return nil
+		},
+	}
+	acctest.TestPlugin(t, testcase)
+}
+
+//go:embed test-fixtures/unlimited-credits/burstable_spot_instances.pkr.hcl
```
lbajolet-hashicorp, line 790: Good idea to leverage embed rather than trying to read it at runtime, we should probably migrate the existing tests that rely on that behaviour to use this instead. But that's for another PR

lbajolet-hashicorp (approved):
Just a nit/question and a comment unrelated to the current code, other than that, LGTM!

builder/common/run_config.go
```diff
@@ -640,6 +663,14 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.RunTags = make(map[string]string)
 	}
 
+	// EnableT2Unlimited has been deprecated so we preserve any config settings.
+	if c.EnableT2Unlimited && !c.EnableUnlimitedCredits {
```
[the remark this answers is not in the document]
lbajolet-hashicorp: OK yeah understood, it's somewhat sad we have to have some duplication here, but it's not a big problem.

We can merge this whenever you're ready, LGTM

PR #278, opened 2022-10-04
Add support for configurating IMDSv2 as the default instance metadata service

### LINKED ISSUE #277: Support IMDSv2 configuration
#### Community Note

#### Description

AWS now supports configuring an AMI to require IMDSv2: https://aws.amazon.com/about-aws/whats-new/2022/10/amazon-machine-images-support-instance-metadata-service-version-2-default/

It would be good if packer could support that (I suspect around https://github.com/hashicorp/packer-plugin-amazon/blob/f1ebbbf409f4863de79a0474ef64d3576faad0e4/builder/instance/step_register_ami.go#L69?).

#### Use Case(s)

We want to enforce IMDSv2 as early and as thoroughly as possible in our environment.

#### Potential configuration

There are already the ["Metadata Settings"](https://www.packer.io/plugins/builders/amazon/ebs#metadata-settings), I guess one could add a flag there?

#### Potential References

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-new-instances.html#configure-IMDS-new-instances-ami-configuration

ankon:
Add support for enabling IMDSv2 by default by using the new `imds_support` configuration option.

Closes #277 

---
Open issues/items:
- [ ] Add examples?
- [ ] Add unit tests?
- [ ] Run acceptance tests

### DIFF
--- a/builder/ebssurrogate/builder_acc_test.go
+++ b/builder/ebssurrogate/builder_acc_test.go
@@ -1,0 +1,129 @@
+package ebssurrogate
+
+import (
+	"fmt"
+	"os/exec"
+	"testing"
+
+	amazon_acc "github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance"
+	"github.com/hashicorp/packer-plugin-sdk/acctest"
+)
+
+func TestAccBuilder_EbssurrogateBasic(t *testing.T) {
+	ami := amazon_acc.AMIHelper{
+		Region: "us-east-1",
+		Name:   "ebs-basic-acc-test",
+	}
+	testCase := &acctest.PluginTestCase{
+		Name:     "amazon-ebssurrogate_basic_test",
+		Template: fmt.Sprintf(testBuilderAccBasic, ami.Name),
+		Teardown: func() error {
+			return ami.CleanUpAmi()
+		},
+		Check: func(buildCommand *exec.Cmd, logfile string) error {
+			if buildCommand.ProcessState != nil {
+				if buildCommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+			return nil
+		},
+	}
+	acctest.TestPlugin(t, testCase)
+}
+
+func TestAccBuilder_EbssurrogateBasic_forceIMDSv2(t *testing.T) {
+	ami := amazon_acc.AMIHelper{
+		Region: "us-east-1",
+		Name:   "ebs-basic-acc-test-imdsv2",
+	}
+	testCase := &acctest.PluginTestCase{
+		Name:     "amazon-ebssurrogate_basic_test_imdsv2",
+		Template: fmt.Sprintf(testBuilderAccBasicIMDSv2, ami.Name),
+		Teardown: func() error {
+			return ami.CleanUpAmi()
+		},
+		Check: func(buildCommand *exec.Cmd, logfile string) error {
+			if buildCommand.ProcessState != nil {
+				if buildCommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+
+			images, err := ami.GetAmi()
+			if err != nil {
+				return fmt.Errorf("failed to get AMI %q: %s", ami.Name, err)
+			}
+
+			if len(images) != 1 {
+				return fmt.Errorf("expected 1 image, got %d", len(images))
+			}
+
+			img := images[0]
+
+			if img.ImdsSupport != nil && *img.ImdsSupport != "v2.0" {
+				return fmt.Errorf("expected AMI to have IMDSv2 support, got %q", *img.ImdsSupport)
+			}
+
+			return nil
+		},
+	}
+	acctest.TestPlugin(t, testCase)
+}
+
+const testBuilderAccBasic = `
+source "amazon-ebssurrogate" "test" {
+	ami_name = "%s"
+	region = "us-east-1"
+	instance_type = "m3.medium"
+	source_ami = "ami-76b2a71e"
+	ssh_username = "ubuntu"
+	launch_block_device_mappings {
+		device_name = "/dev/xvda"
+		delete_on_termination = true
+		volume_size = 8
+		volume_type = "gp2"
+	}
+	ami_virtualization_type = "hvm"
+	ami_root_device {
+		source_device_name = "/dev/xvda"
+		device_name = "/dev/sda1"
+		delete_on_termination = true
+		volume_size = 8
+		volume_type = "gp2"
+	}
+}
+
+build {
+	sources = ["amazon-ebssurrogate.test"]
+}
+`
+
+const testBuilderAccBasicIMDSv2 = `
+source "amazon-ebssurrogate" "test" {
+	ami_name = "%s"
+	region = "us-east-1"
+	instance_type = "m3.medium"
+	source_ami = "ami-76b2a71e"
+	ssh_username = "ubuntu"
+	launch_block_device_mappings {
+		device_name = "/dev/xvda"
+		delete_on_termination = true
+		volume_size = 8
+		volume_type = "gp2"
+	}
+	ami_virtualization_type = "hvm"
+	ami_root_device {
+		source_device_name = "/dev/xvda"
+		device_name = "/dev/sda1"
+		delete_on_termination = true
+		volume_size = 8
+		volume_type = "gp2"
+	}
+	imds_support = "v2.0"
+}
+
+build {
+	sources = ["amazon-ebssurrogate.test"]
+}
+`
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -199,6 +199,11 @@ type Config struct {
 	// more information. Defaults to `legacy-bios` when `ami_architecture` is `x86_64` and
 	// `uefi` when `ami_architecture` is `arm64`.
 	BootMode string `mapstructure:"boot_mode" required:"false"`
+	// Enforce version of the Instance Metadata Service on the built AMI.
+	// Valid options are unset (legacy) and `v2.0`. See the documentation on
+	// [IMDS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html)
+	// for more information. Defaults to legacy.
+	IMDSSupport string `mapstructure:"imds_support" required:"false"`
 
 	ctx interpolate.Context
 }
@@ -502,6 +507,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			AMISkipBuildRegion:       b.config.AMISkipBuildRegion,
 			PollingConfig:            b.config.PollingConfig,
 			BootMode:                 b.config.BootMode,
+			IMDSSupport:              b.config.IMDSSupport,
 		},
 		&awscommon.StepAMIRegionCopy{
 			AccessConfig:      &b.config.AccessConfig,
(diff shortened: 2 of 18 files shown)


ankon:
It's on the "todo list" for me, but I think I missed your reply @nywilken, sorry for that. 

The task items there are a bit questions towards you guys: What are the best practices? Ideally, "where can I clone some code and adapt it", of course. 

As to running acceptance tests: I understand that these might create resources etc, and I haven't had the time to review the various tests to see what kind of cleanups I would need to do and/or how to properly set up an account that doesn't run over costs. I was somewhat hoping that these tests run as part of CI :)

sparrc:
Hello, I'm interested in this feature happy to help out any way I can. @nywilken what are you looking for in terms of "acceptance testing" and "examples"?

I see that @ankon has added docs describing the flag, does that cover "examples"?

lbajolet-hashicorp:
Hi @ankon, sorry for leaving you hanging on this PR, we'll try to get it merged ASAP.

Regarding the examples, as @sparrc mentions, I think the documentation is fine on its own, there are no examples of configuration here for the builders impacted, so I think that point can be left out for this PR.

Unit tests maybe can check that the option can be expressed in the configuration and make sure we do propagate the value. Looking at your code, it seems that although the steps have been updated for all relevant builders (`chroot`, `ebssurrogate` and `instance`), but it looks like only the `ebssurrogate` builder had its config updated, so this option will only be available here, we should update this before we can merge the PR.

For the acceptance tests, right now only `ebs` has some implemented, since this PR changes the code for other builders, there's no existing template for such acceptance testing. This doesn't mean there shouldn't be, but rather that our code is lacking on that front.
Since we have nothing yet, I feel this would be an overreach from us to ask you to implement those from scratch, so I will look into what's to do for those builders.

What I would suggest to move forward on this:

* The configs for the `instance` and `chroot` builders should be updated so their `step_register_ami` can leverage the IMDSv2 setting
* The builder for `ebssurrogate` does not propagate the configuration option you added, I would suggest adding it too when the step is initialised: https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/ebssurrogate/builder.go#L413, otherwise the code will never be used. This code will also need to be updated for the other builders too when their respective config has the option defined.
* The tests for each builder can be updated too, so we can confirm the values are set correctly

I understand that this may be a lot, so please do let me know if you want me to take a jab at this @ankon.

On another note, I understand this IMDSv2 setting is available on the `RegisterImage` API call, hence why `ebssurrogate`, `instance` and `chroot` are the only ones updated, but I wonder if this will impact those that build with `ebs` or `ebsvolume`?

In the meantime, I'll go on and see what we can do for acceptance testing.
This will be done in a further PR, no need to clutter this one with it, and when this is merged, everyone will be able to test the plugin with this extra setting.

Thanks again for opening this PR and the issue to go alongside it, much appreciated!

pisub:
Hi @lbajolet-hashicorp - is there any date when you're expecting to have this PR merged? What's the timeline regarding releasing a new version of plugin including this change?
I'm asking since waiting badly to have this parameter added :)

lbajolet-hashicorp:
Hi @pisub,

That's a good question, I can't commit to a date yet, but ideally we'd merge this sooner than later, since the issue had quite a bit of traction. I'll keep an attentive eye on this.

lbajolet-hashicorp:
Hi @ankon,

I've taken the liberty to push some updates on your branch for this feature, please take a look at it, and let me know if you see something that catches your eye.

@nywilken could you take a look at this also? Since I've modified the code it'd be helpful to have your thoughts.

Independently, since IMDSSupport can only be either `legacy` or `v2.0`, I feel like this could become `enforce_imds_v2`? What do you think? I feel this is slightly clearer in terms of what it implies relative to the produced AMIs.

Thanks again for producing the base code @ankon, hopefully we can merge this fairly soon!

nywilken:
> Independently, since IMDSSupport can only be either `legacy` or `v2.0`, I feel like this could become a `enforce_imds_v2` boolean value? What do you think? I feel this is slightly clearer in terms of what it implies relative to the produced AMIs.

@lbajolet-hashicorp thanks for working with @ankon on pushing this forward . I will take a look at the updates. 
That said, I do agree with you  line of thinking on the attribute name. Instead of `enforce_imds_v2` I would suggest `use_imds_v2`, which is the prefix we tend to use in Packer for these types of boolean attributes.

nywilken:
One thing that does come to mind is if AWS will eventually have a v2.x or v3.0. If so then the current attribute works and doesn't lock us into the situation we had with t2/t3 instances.

nywilken:
@lbajolet-hashicorp quickly looking at the [Go SDK ](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#RegisterImageInput)I see that there is an enum of supported values. We may want to keep the attribute name as is `imds_support` and use the provided enum to validate the possible values. What do you think? 

```
    // Set to v2.0 to indicate that IMDSv2 is specified in the AMI. Instances launched
    // from this AMI will have HttpTokens automatically set to required so that,
    // by default, the instance requires that IMDSv2 is used when requesting instance
    // metadata. In addition, HttpPutResponseHopLimit is set to 2. For more information,
    // see Configure the AMI (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-IMDS-new-instances.html#configure-IMDS-new-instances-ami-configuration)
    // in the Amazon Elastic Compute Cloud User Guide.
    //
    // If you set the value to v2.0, make sure that your AMI software can support
    // IMDSv2.
    ImdsSupport *string `type:"string" enum:"ImdsSupportValues"`
```

lbajolet-hashicorp:
I agree on the usual name when it describes some information at build-time, however from my understanding, this attribute is used to enforce IMDSv2 on the resulting AMI, so for this specific case, I would still think `enforce_imds_v2` would be more appropriate.

I'm open to other suggestion however :)

Your point on future-proofness does stand however, and indeed then, the current approach does not lock us in in case that changes at some point later in the future.
We can keep it as-is for the time being if future-proofness is the main concern, the current approach is satisfactory in my opinion, albeit a bit verbose for my liking, hence the boolean flag proposal, but it does work well enough, and does not lock us for further updates.

If we do decide on keeping the option as defined, I would then add some extra checks during the `Prepare` step for all those builders, so we ensure the value is accepted before we proceed with the build itself, otherwise we run the risk of finishing the build, and the register-ami step will fail with an unsupported value.

I'll send an update later to that PR.

ankon:
Thanks a lot for going ahead here, @lbajolet-hashicorp! 

I had a quick look, but basically this codebase is completely unknown to me, and I don't think I can say much about it at all -- I'm already happy that it looks like I wasn't completely off the mark :)

As to the naming: I also thought about whether "use/enable support v2" or "configure version" was the right way of thinking about this, and ultimately went the latter option because indeed AWS seems to have everything in place for not just a simple boolean flag, but rather a complete versioning infrastructure.

lbajolet-hashicorp:
Hey there,

FYI: the PR has been updated with the extra value checks in the Prepare step of the builder, by that point, I think we can drop the WIP status for the PR, and consider merging it if we're happy with the latest version of the code.

@nywilken when you have time if you can do a last review, and if all aligns here, we can merge this.

builder/ebssurrogate/builder_acc_test.go
```diff
@@ -0,0 +1,135 @@
...
+	"time"
+
+	amazon_acc "github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance"
+	"github.com/hashicorp/packer-plugin-sdk/acctest"
+)
+
+func TestAccBuilder_EbssurrogateBasic(t *testing.T) {
+	ami := amazon_acc.AMIHelper{
+		Region: "us-east-1",
+		Name:   fmt.Sprintf("packer-plugin-amazon-ebs-basic-acc-test %d", time.Now().Unix()),
```
nywilken, line 16: Depending on how fast these test cases run there is no guarantee  that these AMI names will be different. I suggest giving them unique names to avoid any potential naming conflicts
```suggestion
		Name:   "ebsurrogate-basic-acc-test",
```

builder/ebssurrogate/builder_acc_test.go
```diff
@@ -0,0 +1,135 @@
...
+				}
+			}
+			return nil
+		},
+	}
+	acctest.TestPlugin(t, testCase)
+}
+
+func TestAccBuilder_EbssurrogateBasic_forceIMDSv2(t *testing.T) {
+	aminame := fmt.Sprintf("packer-plugin-amazon-ebs-basic-acc-test %d", time.Now().Unix())
```
nywilken, line 37: I suggest giving making this unique as well to avoid any potential naming conflicts
```suggestion
	aminame := "ebsurrogate-acc-test-imdsv2"
```

builder/ebssurrogate/builder_acc_test.go
```diff
@@ -0,0 +1,135 @@
...
+			if buildCommand.ProcessState != nil {
+				if buildCommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+
+			ami := amazon_acc.AMIHelper{
+				Region: "us-east-1",
+				Name:   aminame,
+			}
```
nywilken, line 58: You make this call on line 38, as it's needed in the Teardown function. Can use reuse ami here? Or does you need to reinitialize?

nywilken (approved):
This looks great! Nicely done on adding additional acceptance testing. I left a comment on the ami name for testing, which is failing locally for me. But other than that this looks good to me. I'm approving in advance so you can merge when ready. 

```
==> Some builds didn't complete successfully and had errors:
2022/12/01 10:37:08 machine readable: amazon-ebssurrogate.test,error []string{"Error registering AMI: InvalidAMIName.Duplicate: AMI name packer-plugin-amazon-ebs-basic-acc-test 1669908903 is already in use by AMI ami-00789d0cf6c335352\n\tstatus code: 400, request id: 997475e7-ec0c-4def-b152-669541cca106"}
2022/12/01 10:37:08 ui error: --> amazon-ebssurrogate.test: Error registering AMI: InvalidAMIName.Duplicate: AMI name packer-plugin-amazon-ebs-basic-acc-test 1669908903 is already in use by AMI ami-00789d0cf6c335352
	status code: 400, request id: 997475e7-ec0c-4def-b152-669541cca106
```

builder/ebssurrogate/builder_acc_test.go
lbajolet-hashicorp: I'm not sure what you mean by "reuse the ami here", could you clarify? Both tests have their own AMI with a different name after integrating your last suggestions, anything else that you think could be relevant regarding those tests?

lbajolet-hashicorp: OK my bad, I just noticed the duplicate. I've removed it, rerolling the commit

pisub:
So awesome to see this being merged, you guys rock!
@lbajolet-hashicorp is there any new info on release date of plugin including this PR?

lbajolet-hashicorp:
Hi @pisub,

The plan is to release a new version of the plugin this week, probably today or tomorrow!

robin-fordham:
Hi all,

Great work!

Is there are reason why this functionality is not included on the ebs builder? and what is a suggested route for those using the ebs builder?

Thanks in advance!

lbajolet-hashicorp:
Hi @robin-fordham,

The only reason why this is not on the `ebs` builder is because AMIs are registered differently for this case, and the `imds_support` option for the AMI is only (AFAIK) on the `RegisterImage` API call, so that's why it wasn't included in this PR.

We could probably rework the way we register our images for the EBS builder so this functionality becomes available, or we can wait until AWS offers this possibility in the API endpoints we use for the ebs builder (CreateImage).

Given how many users rely on the current way we create the AMI for the ebs builder, I would be very cautious changing it, since it may break a lot of existing workflows, typically because of permission issues or because we risk introducing new bugs during this refactoring phase. I imagine if this option gains traction on the community side, Amazon will likely introduce this option for other API calls, so this seems like our best bet if this is not too long.

Sorry for not bringing better news, hopefully this problem will be sorted out soon.

jddemcher:
If imds_support is not yet available for the ebs builder, what is the purpose of the metadata-options and http_tokens flags? Could those not be used to accomplish the same result?

robin-fordham:
@jddemcher those options are in the **run configuration** portion of the builder, my understanding is they are configuration options for the EC2 instance that is spun up to perform the build, rather than on option for the resulting AMI.

Hope this helps.

PR #282, opened 2022-10-18
Update RunConfig-not-required.mdx

EdenBW:
'Resource' field should be list of strings as per [builder/common/run_config.go](https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/common/run_config.go#L42)


lbajolet-hashicorp (approved):
Good catch!

Thanks for the fix @EdenBW, merging this now

PR #293, opened 2022-11-16
Remove docs preview from make generate target

nywilken:


nywilken:
Merging change to remove docs preview.

PR #296, opened 2022-11-23
session_manager: Upload user public key to instance before connection if private key is specified

### LINKED ISSUE #31: SSM Session Manager Error when user has not access to create keypair in AWS
_This issue was originally opened by @mixeract as hashicorp/packer#10453. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

Our AWS accounts have a policy preventing roles from creating SSH keypairs. This prevents Packer from generating temporary key. If we don't specify a key in the configs packer will automatically assumes that automatic key generation is required. This will fail right away as the user has no permission to create key pairs. 

```
amazon-ebs: Creating temporary keypair: packer_5eb99fdc-a11e-d9a0-e429-356b343ccc69 Build 'amazon-ebs' errored: Error creating temporary keypair: retry count exhausted. Last err: UnauthorizedOperation: You are not authorized to perform this operation.
```

In order to prevent packer from auto generate the keys, I've declared a keypair in the configs `"ssh_private_key_file": "~/.ssh/packer_key"`, however by doing that Packer is ignoring the ssm session manager and tries to create ssh session using the provided key.

Error will be 
```
==> amazon-ebs: Error waiting for SSH: Packer experienced an authentication error when trying to connect via SSH. This can happen if your username/password are wrong. You may want to double-check your credentials as part of your debugging process. original error: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
==> amazon-ebs: Terminating the source AWS instance...
```
Configs 
```
  "builders": [
    {
      "type": "amazon-ebs",
      "ami_name": "{{ user `name` }}",
      "profile": "{{ user `profile` }}",
      "source_ami_filter": {
        "filters": {
          "virtualization-type": "hvm",
          "name": "internal_ssm_ami*",
          "root-device-type": "ebs"
        },
        "owners": "ouraccountnumber",
        "most_recent": true
      },
      "ssh_username": "ec2-user",
      "ssh_pty": true,
      "pause_before_ssm": "1m",
      "ssh_interface": "session_manager",
      "communicator": "ssh",
      "ssh_timeout": "5m",
      "ssh_private_key_file": "~/.ssh/packer_key",
      "iam_instance_profile": "packer_instance_profile",
      "security_group_ids" : [list_of_sg],
      "disable_stop_instance": true,
      "shutdown_behavior": "stop",
      "session_manager_port" : 8422,
      "instance_type": "{{ user `instance_type` }}",
      "subnet_id": "{{ user `subnet_id` }}",
      "vpc_id": "{{ user `vpc_id` }}",
      "launch_block_device_mappings": [
        {
          "device_name": "/dev/xvda",
          "volume_size": 12,
          "volume_type": "gp2",
          "delete_on_termination": true
        }
      ],
    }
  ],
```

Any idea if there's a solution to this? Maybe an option that I missed In the docs prevents Packer from using the private key or a config to prevent packer from auto-generation of keys?

Glyphack:
Improves SSM session manager feature by disabling key pair creation when user specifies a private key pair to be used.

The current implementation works fine for cases where the user does not specify a private key to be used with `session_manager`. In cases where the user specifies a private key, then that key will be used to authenticate but it fails since the key is not present on the EC2 instance.

This PR will resolve the issue by [sending the user public key](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2instanceconnect/#EC2InstanceConnect.SendSSHPublicKey) to the instance before connecting to it.

## The New Behavior

If the user does not specify a private key, then works like the previous implementations.

In the case where the user specifies a private key to be used:
1. key pair will not be created.
2. the public key will be sent to the instance before connecting to it. This key will live for 60 seconds which is enough for the handshake.

## Questions

There are some limitations to what public keys we can send to instances with SSM.
The SendSSHPublicKeyInput struct in aws-sdk-go currently used in the project only supports RSA keys because it has a limit on the length of the public key.
This behavior [is fixed](https://github.com/aws/aws-sdk-go/blob/main/service/ec2instanceconnect/api.go?rgh-link-date=2022-12-03T13%3A11%3A40Z#L651) in the recent version so if we update it then it supports other keys.

How can I update the docs?

I'll complete the code and add it to other builders once I get initial feedback on the implementation.

Closes #31

### DIFF
--- a/builder/common/step_create_ssm_tunnel.go
+++ b/builder/common/step_create_ssm_tunnel.go
@@ -3,13 +3,18 @@ package common
 import (
 	"context"
 	"fmt"
+	"log"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2instanceconnect"
 	"github.com/aws/aws-sdk-go/service/ssm"
 	pssm "github.com/hashicorp/packer-plugin-amazon/builder/common/ssm"
+	"github.com/hashicorp/packer-plugin-sdk/communicator"
+	"github.com/hashicorp/packer-plugin-sdk/communicator/sshkey"
 	"github.com/hashicorp/packer-plugin-sdk/multistep"
 	"github.com/hashicorp/packer-plugin-sdk/net"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
@@ -21,6 +26,7 @@ type StepCreateSSMTunnel struct {
 	LocalPortNumber  int
 	RemotePortNumber int
 	SSMAgentEnabled  bool
+	SSHConfig        *communicator.SSH
 	PauseBeforeSSM   time.Duration
 	stopSSMCommand   func()
 }
@@ -81,7 +87,47 @@ func (s *StepCreateSSMTunnel) Run(ctx context.Context, state multistep.StateBag)
 		}
 	}()
 
+	if len(s.SSHConfig.SSHPrivateKey) != 0 {
+		ui.Say("Uploading SSH public key to instance")
+		err := s.sendUserSSHPublicKey(instance, s.SSHConfig.SSHPrivateKey)
+		if err != nil {
+			ui.Error(err.Error())
+			return multistep.ActionHalt
+		}
+	}
+
 	return multistep.ActionContinue
+}
+
+func (s *StepCreateSSMTunnel) sendUserSSHPublicKey(
+	instance *ec2.Instance,
+	privateKey []byte,
+) error {
+	publicKey, err := sshkey.PublicKeyFromPrivate(privateKey)
+	if err != nil {
+		return fmt.Errorf("Error getting public key from private key: %s", err)
+	}
+	svc := ec2instanceconnect.New(s.AWSSession)
+	input := &ec2instanceconnect.SendSSHPublicKeyInput{
+		AvailabilityZone: aws.String(*instance.Placement.AvailabilityZone),
+		InstanceId:       aws.String(*instance.InstanceId),
+		InstanceOSUser:   aws.String(s.SSHConfig.SSHUsername),
+		SSHPublicKey:     aws.String(strings.TrimSuffix(string(publicKey), "\n")),
+	}
+	log.Printf("Sending public key to instance: %s", *input.InstanceId)
+	result, err := svc.SendSSHPublicKey(input)
+	if err != nil {
+		err := fmt.Errorf(`
+        error encountered in sending public key to instance: %s
+      Check the key type and length are valid in AWS API.
+      https://docs.aws.amazon.com/ec2-instance-connect/latest/APIReference/API_SendSSHPublicKey.html`, err)
+		return err
+	} else {
+		if *result.Success {
+			return nil
+		}
+	}
+	return fmt.Errorf("Failed to send public key to instance")
 }
 
 // Cleanup terminates an active session on AWS, which in turn terminates the associated tunnel process running on the local machine.
--- a/builder/ebs/acceptance/utils.go
+++ b/builder/ebs/acceptance/utils.go
@@ -1,0 +1,22 @@
+package amazon_acc
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func GenerateSSHPrivateKeyFile() (string, error) {
+	outFile := fmt.Sprintf("%s/temp_key.ed25519", os.TempDir())
+	sshGenCmd := exec.Command("ssh-keygen", "-t", "ed25519", "-b", "256", "-f", outFile)
+	sshGenCmd.Stdin = bytes.NewBuffer([]byte("\n\n"))
+
+	_, err := sshGenCmd.CombinedOutput()
+	if err != nil {
+		os.Remove(outFile)
+		return "", err
+	}
+
+	return outFile, nil
+}
--- a/builder/ebssurrogate/builder.go
+++ b/builder/ebssurrogate/builder.go
@@ -389,6 +389,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			LocalPortNumber:  b.config.SessionManagerPort,
 			RemotePortNumber: b.config.Comm.Port(),
 			SSMAgentEnabled:  b.config.SSMAgentEnabled(),
+			SSHConfig:        &b.config.Comm.SSH,
 		},
 		&communicator.StepConnect{
 			Config: &b.config.RunConfig.Comm,
--- a/builder/ebs/builder.go
+++ b/builder/ebs/builder.go
@@ -339,6 +339,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			LocalPortNumber:  b.config.SessionManagerPort,
 			RemotePortNumber: b.config.Comm.Port(),
 			SSMAgentEnabled:  b.config.SSMAgentEnabled(),
+			SSHConfig:        &b.config.Comm.SSH,
 		},
 		&communicator.StepConnect{
 			Config: &b.config.RunConfig.Comm,
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -771,6 +771,12 @@ func TestAccBuilder_EbsKeyPair_rsaSHA2OnlyServer(t *testing.T) {
 }
 
 func TestAccBuilder_PrivateKeyFile(t *testing.T) {
+	if os.Getenv(acctest.TestEnvVar) == "" {
+		t.Skipf("Acceptance tests skipped unless env '%s' set",
+			acctest.TestEnvVar)
+		return
+	}
+
 	ami := amazon_acc.AMIHelper{
 		Region: "us-east-1",
 		Name:   fmt.Sprintf("packer-pkey-file-acc-test-%d", time.Now().Unix()),
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -770,6 +770,33 @@ func TestAccBuilder_EbsKeyPair_rsaSHA2OnlyServer(t *testing.T) {
 	acctest.TestPlugin(t, testcase)
 }
 
+func TestAccBuilder_PrivateKeyFile(t *testing.T) {
+	ami := amazon_acc.AMIHelper{
+		Region: "us-east-1",
+		Name:   fmt.Sprintf("packer-pkey-file-acc-test-%d", time.Now().Unix()),
+	}
+
+	sshFile, err := amazon_acc.GenerateSSHPrivateKeyFile()
+	if err != nil {
+		t.Fatalf("failed to generate SSH key file: %s", err)
+	}
+
+	defer os.Remove(sshFile)
+
+	testcase := &acctest.PluginTestCase{
+		Name:     "amazon-ebs_test_private_key_file",
+		Template: buildPrivateKeyFileConfig(ami.Name, sshFile),
+		Check: func(buildCommand *exec.Cmd, logfile string) error {
+			if buildCommand.ProcessState.ExitCode() != 0 {
+				return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+			}
+			return nil
+		},
+	}
+
+	acctest.TestPlugin(t, testcase)
+}
+
 //go:embed test-fixtures/unlimited-credits/burstable_instances.pkr.hcl
 var testBurstableInstanceTypes string
 
@@ -986,6 +1013,24 @@ const testBuilderAccEnableDeprecation = `
 }
 `
 
+const testPrivateKeyFile = `
+source "amazon-ebs" "test" {
+	ami_name             = "%s"
+	source_ami           = "ami-0b5eea76982371e91" # Amazon Linux 2 AMI - kernel 5.10
+	instance_type        = "m3.medium"
+	region               = "us-east-1"
+	ssh_username         = "ec2-user"
+	ssh_interface        = "session_manager"
+	iam_instance_profile = "SSMInstanceProfile"
+	communicator         = "ssh"
+	ssh_private_key_file = "%s"
+}
+
+build {
+	sources = ["amazon-ebs.test"]
+}
+`
+
 func buildForceDeregisterConfig(val, name string) string {
 	return fmt.Sprintf(testBuilderAccForceDeregister, val, name)
 }
@@ -1000,4 +1045,8 @@ func buildSharingConfig(val, val2, val3, name string) string {
 
 func buildEnableDeprecationConfig(val, name string) string {
 	return fmt.Sprintf(testBuilderAccEnableDeprecation, val, name)
+}
+
+func buildPrivateKeyFileConfig(name, keyPath string) string {
+	return fmt.Sprintf(testPrivateKeyFile, name, keyPath)
 }
--- a/builder/ebssurrogate/builder_acc_test.go
+++ b/builder/ebssurrogate/builder_acc_test.go
@@ -2,8 +2,10 @@ package ebssurrogate
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"testing"
+	"time"
 
 	amazon_acc "github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance"
 	"github.com/hashicorp/packer-plugin-sdk/acctest"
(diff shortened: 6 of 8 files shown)


Glyphack:
I tried to connect manually the way packer connects and seems like the issue is that when there is no key pair then ssh connection is denied because of public key.
So I did a sent the ssh key before the ssh and it worked.
https://docs.aws.amazon.com/sdk-for-go/api/service/ec2instanceconnect/#EC2InstanceConnect.SendSSHPublicKey
Maybe this action needs to done before connect step.

Glyphack:
This is the working process I came up with:
1. Users must specify a private ssh key in their packer file. We will use this to authenticate over SSH.
2. Before starting the SSH port forwarding [We send the public key](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2instanceconnect/#EC2InstanceConnect.SendSSHPublicKey) that user specified to the instance.

The rest of the process is the same. I'll see if it's possible to continue with a generated key instead of asking the user for a key.

Glyphack:
@nywilken Since you implemented this feature originally, can you give me some feedback on this approach?

lbajolet-hashicorp (changes requested):
Hey @Glyphack, sorry for the delay in reviewing this PR.

The code overall looks good to me, I was worried that we could timeout waiting for the instance to be up, but it looks like we wait for the instance to be up, so that 60s delay should suffice in between when we send the key and when we attempt to connect, so we should be good to go.

I would maybe consider adding an acceptance test to make sure this works as intended, but overall this is fine.

I left a few comments on the PR as well that need to be addressed, after that we can consider merging.

Thanks for the fix!

builder/common/step_create_ssm_tunnel.go
```diff
@@ -75,15 +81,51 @@ func (s *StepCreateSSMTunnel) Run(ctx context.Context, state multistep.StateBag)
...
+			return multistep.ActionHalt
+		}
+		ui.Say(fmt.Sprintf("Uploaded public key to instance"))
+	}
+
 	return multistep.ActionContinue
 }
 
+func (s *StepCreateSSMTunnel) sendUserSSHPublicKey(instance *ec2.Instance, privateKey []byte) error {
+	publicKeyBytes, err := sshkey.PublicKeyFromPrivate(privateKey)
```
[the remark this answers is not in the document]
Glyphack: I think the SSHPublic key is never set. I'll try it out.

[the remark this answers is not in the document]
Glyphack: [This](https://github.com/hashicorp/packer-plugin-sdk/blob/main/communicator/step_ssh_keygen.go#L43) is the part that set the value for SSHPublicKey. But somehow the value is not there when I try to access it.

Could you check [this commit](https://github.com/hashicorp/packer-plugin-amazon/pull/296/commits/e5e6ce415c4e5efd72bb6acec50093c9dda09574) of how I'm doing this, maybe I'm making a mistake?

Glyphack:
Thanks for the review @lbajolet-hashicorp. I resolved the comments, but I would appreciate your input on the remaining topics:
https://github.com/hashicorp/packer-plugin-amazon/pull/296#discussion_r1053373320

Also I'm not familiar with how the acceptance tests work, could you help me with how to write one?
This is the configuration required to test this bugfix:

```
source "amazon-ebs" "hvr" {
  ami_name      = "test-${formatdate("YYYY-MM-DD-hhmmss", timestamp())}"
  instance_type = "t2.micro"
  region        = "eu-west-1"
  source_ami_filter {
    filters = {
      name                = "amzn2-ami-hvm-*-x86_64-ebs"
      root-device-type    = "ebs"
      virtualization-type = "hvm"
    }
    most_recent = true
    owners      = ["amazon"]
  }
  ssh_username  = "ec2-user"
  ssh_interface = "session_manager"
  communicator  = "ssh"
  temporary_iam_instance_profile_policy_document {
    Statement {
      Action   = ["sts:AssumeRole"]
      Effect   = "Allow"
      Resource = ["*"]
      Principal = {
        "Service" : "ec2.amazonaws.com"
      }
    }
    Version = "2012-10-17"
  }
  ssh_private_key_file = "~/.ssh/id_rsa"
  vpc_id               = "vpc-0d2aa27cfe71c1bc4"
  subnet_id            = "subnet-0e4b7f6537867c07b"
}

build {
  name = "hvr"
  sources = [
    "source.amazon-ebs.hvr"
  ]
}
```

builder/common/step_create_ssm_tunnel.go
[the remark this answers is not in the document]
lbajolet-hashicorp: Hm, looks like the PublicKey is only set when we generate an SSH key, my bad here! Though tbf maybe both should be set when we read a private key from the outside. I'll take a look at the SDK, this could be fixed here for a later release.

In the meantime, I imagine your initial implementation would work, sorry for misleading you on there.

Also, come to think of it, it feels like this should also work when the key is read from a file (see the `SSHPrivateKeyFile` attribute from the `SSHConfig` struct). Unless I'm mistaken, this was not supported in your original implementation, correct?

There's also a `SSHCertificateFile` that may be relevant there too, what do you think?

Considering the multiple possibilities here, the whole `Communicator.SSH` can be passed along, and we can then choose which to upload (or maybe we can upload all the relevant public keys, if that's a possibility?).

lbajolet-hashicorp:
Hi @Glyphack,

Just wanted to ping on this PR, do you need assistance on landing this?
I can probably set some time to work on it next week if you need help, let me know.

Glyphack:
Hey @lbajolet-hashicorp thanks
The last comment is clear to me; I'm going to apply it.
I can get some help with [writing the acceptance](https://github.com/hashicorp/packer-plugin-amazon/pull/296#issuecomment-1364724264) test in case it's needed.

builder/common/step_create_ssm_tunnel.go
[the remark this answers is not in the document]
Glyphack: Yeah I was originally testing with providing the key via file. since I think this issue only happens when you specify the file, otherwise packer/amazon will take care of adding the key to instance.

But I don't this hurts to use the SSHConfig and upload all the public keys that are used in the session to the instance. Also this issue might happen for other cases too, and I was only fixing it for the problem I knew.

The `SSHCertificateFile` seems relevant. I did not know Go had these constructs for working with SSH. I we add an `AuthMethod` from the certs and probably I can scan all the public key auth methods and add them to the instance. This way it works in a more generic way.

Thanks for the suggestion. Will apply & test it tomorrow.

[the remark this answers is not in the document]
Glyphack: So I'm going with your suggestion, passing the `Communicator.SSH` and uploading the public keys I mentioned above.

lbajolet-hashicorp:
Hi @Glyphack,

Sure thing, when you're done with the latest reroll, let me know, and I'll take a jab at the acceptance testing.

For reference, acceptance tests are written for `ec2` and `ebssurrogate` for now (SEE: https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/ebs/builder_acc_test.go for example).

When we're ready to test the feature, I'll get to work on that.

Thanks again!

builder/common/step_create_ssm_tunnel.go
```diff
@@ -21,6 +25,8 @@ type StepCreateSSMTunnel struct {
 	LocalPortNumber  int
 	RemotePortNumber int
 	SSMAgentEnabled  bool
+	SSHUsername      string
+	Comm             *communicator.Config
```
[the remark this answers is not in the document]
Glyphack: As [discussed](https://github.com/hashicorp/packer-plugin-amazon/pull/296?notification_referrer_id=NT_kwDOAT00brM0OTA5MDY2MTg0OjIwNzg4MzM0#discussion_r1062900776) here we're sending the whole SSHConfig

Glyphack:
Hi @lbajolet-hashicorp 

This is ready for review and test(I did it locally). Just a few notes/questions
This is the final packer file I used to test the functionality. It is based on the original issue:
```
source "amazon-ebs" "hvr" {
  ami_name      = "test-${formatdate("YYYY-MM-DD-hhmmss", timestamp())}"
  instance_type = "t2.micro"
  region        = "eu-west-1"
  source_ami_filter {
    filters = {
      name                = "amzn2-ami-hvm-*-x86_64-ebs"
      root-device-type    = "ebs"
      virtualization-type = "hvm"
    }
    most_recent = true
    owners      = ["amazon"]
  }
  ssh_username  = "ec2-user"
  ssh_interface = "session_manager"
  communicator  = "ssh"
  /* This is just a profile with SSMManagedInstanceCore policy attached */
  iam_instance_profile = "packer_profile"
  /* This is my private key which is used for all ssh connections by default */
  ssh_private_key_file = "~/.ssh/id_ed25519"
```

I have also added this to other builders as well(initially it was only working with ebs). I haven't used the other builders so if something is not compatible with this change let me know.

builder/common/step_create_ssm_tunnel.go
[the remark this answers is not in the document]
Glyphack: @lbajolet-hashicorp I tried to only pass the `Communicator.SSH` but there's a method on the `Communicator.Config` which is useful here:
```
CommunicatorConfig.ReadSSHPrivateKeyFile()
```

I'm reading the private key file with this method.

builder/ebssurrogate/builder.go
```diff
@@ -66,15 +66,15 @@ type Config struct {
...
 	// what architecture to use when registering the final AMI; valid options
 	// are "arm64", "i386", "x86_64", or "x86_64_mac". Defaults to "x86_64".
-	Architecture string `mapstructure:"ami_architecture" required:"false"`
+	Architecture string `mapstructure:"ami_architecture"             required:"false"`
 	// The boot mode. Valid options are `legacy-bios` and `uefi`. See the documentation on
 	// [boot modes](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ami-boot.html) for
 	// more information. Defaults to `legacy-bios` when `ami_architecture` is `x86_64` and
 	// `uefi` when `ami_architecture` is `arm64`.
-	BootMode string `mapstructure:"boot_mode" required:"false"`
+	BootMode string `mapstructure:"boot_mode"                    required:"false"`
```
lbajolet-hashicorp, line 77: I wonder, is this a linter's doing? I'm not sure I understand what's this modification for, since it adds a bunch of spaces to almost all the config values, without clear benefit.
I'm not necessarily opposed to lint changes, but in this case I find this one distracting, could I ask you to remove those from the changes in the PR?

builder/common/step_create_ssm_tunnel.go
```diff
@@ -75,15 +84,70 @@ func (s *StepCreateSSMTunnel) Run(ctx context.Context, state multistep.StateBag)
...
 			Region:     s.Region,
 		}.Start(ssmCtx, ui)
-
 		if err != nil {
 			ui.Error(fmt.Sprintf("ssm error: %s", err))
 		}
 	}()
 
+	var privateKeys [][]byte
+	if len(s.CommunicatorConfig.SSH.SSHPrivateKey) != 0 {
```
lbajolet-hashicorp, line 93: I'd think we can replace the length check by an empty string comparison for consistency with the `SSHPrivateKeyFile` check below. That is unless I'm misunderstanding the reasons here, if there's a good reason for this implementation we can leave it as-is.
```suggestion
	if s.CommunicatorConfig.SSH.SSHPrivateKey != "" {
```

builder/common/step_create_ssm_tunnel.go
```diff
@@ -75,15 +84,70 @@ func (s *StepCreateSSMTunnel) Run(ctx context.Context, state multistep.StateBag)
...
+	}
+	log.Printf("public key: %s", publicKey)
+	svc := ec2instanceconnect.New(s.AWSSession)
+	input := &ec2instanceconnect.SendSSHPublicKeyInput{
+		AvailabilityZone: aws.String(*instance.Placement.AvailabilityZone),
+		InstanceId:       aws.String(*instance.InstanceId),
+		InstanceOSUser:   aws.String(s.CommunicatorConfig.SSH.SSHUsername),
+		SSHPublicKey:     aws.String(strings.TrimSuffix(string(publicKey), "\n")),
+	}
+	log.Printf("Sending public key to instance: %s", input)
```
lbajolet-hashicorp, line 135: This log will probably be verbose, I suggest we either cherry pick the instance ID (or anything that makes sense here) and only print this out

builder/common/step_create_ssm_tunnel.go
```diff
@@ -75,15 +84,70 @@ func (s *StepCreateSSMTunnel) Run(ctx context.Context, state multistep.StateBag)
...
 
+func (s *StepCreateSSMTunnel) sendUserSSHPublicKey(
+	instance *ec2.Instance,
+	privateKey []byte,
+) error {
+	publicKey, err := sshkey.PublicKeyFromPrivate(privateKey)
+	if err != nil {
+		return fmt.Errorf("Error getting public key from private key: %s", err)
+	}
+	log.Printf("public key: %s", publicKey)
```
lbajolet-hashicorp, line 127: Regarding this log I don't think it provides meaningful information, we can probably safely remove it.

lbajolet-hashicorp (commented):
Hi @Glyphack,

I've taken another look at the code, I left a few comments on it which can be addressed when you are available, no rush here.

I'll write some acceptance tests for this change, and when we're good on this, I figure this can be merged.

Thanks for the rework, I think it's good that we support both the file and literal key for this.

builder/common/step_create_ssm_tunnel.go
lbajolet-hashicorp: Well I didn't see this before, but it seems that if `SSHPrivateKeyFile` is defined in the configuration, it will set `SSHPrivateKey` to that value (SEE: https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/common/step_key_pair.go#L34).

Therefore I think we can remove the `if s.CommunicatorConfig.SSH.SSHPrivateKeyFile != ""` block, since it's redundant with this one (in my testing I can confirm that the key is indeed set twice, while not harmful, it's not necessary)

lbajolet-hashicorp (approved):
FYI @Glyphack,

I have written an acceptance test for EBS, I can also write one for ebssurrogate since this is the other builder we have a framework for, but in its current state, I can confirm your code does work well!

Once my latest comments are addressed, I think we're ready to merge this.

Thanks for the fix, looking forward to press the button!

lbajolet-hashicorp:
Oh, there are some conflicts also to fix, I'd suggest rebasing this PR on top of the current main, fixing the conflicts, and pushing the branch again so we're clear to merge

builder/common/step_create_ssm_tunnel.go
Glyphack: Yes exactly, I tested it and the key is set twice. I just wanted to keep this as same of how the [AuthMethod](https://github.dev/hashicorp/packer-plugin-sdk/blob/34ed5d90315ee53e5ebdea1a7c2a68a95b7812b6/communicator/config.go#L394-L395) is set(indeed there it's set twice). But if you agree let's remove it.

lbajolet-hashicorp: That's interesting, I suspect this is an unintended side-effect of setting the PrivateKey as the contents of the SSHPrivateKeyFile.
There's probably something to do on that side, in the meantime, we can proceed with removing the `SSHPrivateKeyFile` condition in this code, it doesn't harm to set the public key twice as far as I can tell, but we might as well do it right if we're on it

builder/common/step_create_ssm_tunnel.go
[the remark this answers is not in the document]
Glyphack: Since we decided to remove the SSHPrivateKeyFile I'm falling back to only passing the `Communicator.SSH`.

builder/common/step_create_ssm_tunnel.go
lbajolet-hashicorp: Oh! That makes sense, sorry I didn't catch that, it's good then!

Glyphack:
@lbajolet-hashicorp Thank you so much for the effort.

I rebased the branch and applied the suggestions.
**Note** I removed the AWS SDK version bump from changes since it's already updated in the main.

lbajolet-hashicorp:
@Glyphack thanks to you for the rerolls and the preseverance!

I took the liberty of pushing a few acceptance tests for both the ebs and the ebssurrogate builders for this feature, I'll see that someone else on the team reviews your PR one last time to make sure I didn't introduce something that I shouldn't, but after that, we can merge it.

Thanks again for the collaboration, much appreciated

builder/ebs/acceptance/utils.go
```diff
@@ -0,0 +1,22 @@
+package amazon_acc
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"os/exec"
+)
+
+func GenerateSSHPrivateKeyFile() (string, error) {
```
JenGoldstrich, line 10: chatted about moving this out of the utils file cause of failures with the windows acceptance test with @lbajolet-hashicorp in a call

JenGoldstrich (approved):

builder/ebs/acceptance/utils.go
lbajolet-hashicorp: For visibility, we're not moving this away, but we add extra checks to avoid running this unless we're running the acceptance tests here.
Hopefully, the last commit should fix this issue.

danielkza:
This breaks builds for setups that were previously self-managing the public key installation and don't have `ec2-instance-connect:SendSSHPublicKey` allowed in IAM policies. The need for this action is also not documented anywhere.

PR #308, opened 2023-01-14
Fix spot instance launch readiness

### LINKED ISSUE #40: spot instances on p3.* fails to connect because packer tries to connect on private IP
_This issue was originally opened by @amrragab8080 as hashicorp/packer#10347. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Overview of the Issue

Launching a p3.* build node with spot instances. The node launches in a public subnet with a public IP, however packer tries to connect on the private IPv4
```
==> amazon-ebs: Using ssh communicator to connect: 172.31.2.186
--
984 | ==> amazon-ebs: Waiting for SSH to become available...
985 | ==> amazon-ebs: Timeout waiting for SSH.
986 | ==> amazon-ebs: Terminating the source AWS instance...
```
Time's out on the connect. However my environment is flexible enough that my packer launch nodes are also on AWS and can reach the build nodes on its private IP. But wanted to report the bug none the less.

#### Reproduction Steps

```
"builders": [{
    "type": "amazon-ebs",
    "region": "{{user `region`}}",
    "source_ami": "{{user `build_ami`}}",
    "run_tags": {
        "Name": "packer-gpu-processor-{{user `flag`}}"
    },
    "subnet_id": "{{user `subnet_id`}}",
    "security_group_ids": "{{user `security_groupids`}}",
    "spot_instance_types": ["p3.8xlarge","p3.16xlarge"],
    "spot_price": "auto",
    "ssh_username": "ubuntu",
    "ami_name": "ml-gpu-processor_{{user `flag`}}-{{timestamp}}",
    "launch_block_device_mappings":[{
      "delete_on_termination": true,
      "device_name": "/dev/sda1",
      "volume_size": 150,
      "volume_type": "gp2"
    }]

```
### Packer version

1.6.5

OS, Architecture, and any other information you can provide about the
environment.

Ubuntu 20.04 local build environment.

Include appropriate log fragments. If the log is longer than a few dozen lines,
please include the URL to the [gist](https://gist.github.com/) of the log or
use the [Github detailed format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) instead of posting it directly in the issue.

Set the env var `PACKER_LOG=1` for maximum log detail.

sparrc:
When launching ec2 source instances, we have code which waits for the instance to become ready, but previously this codepath was not being applied to spot instances.

This change factors that code out into a common function, and then calls that common function in both the 'regular' and 'spot' instance launches.

This fixes issues where packer tries to connect to a spot instance before it is running and ready, which means it won't have a public ip address ready yet, and so packer times out on SSH access.

fixes #40

### DIFF
--- a/builder/common/state.go
+++ b/builder/common/state.go
@@ -45,18 +45,22 @@ type StateChangeConf struct {
 //
 // HCL2 example:
 // ```hcl
-// aws_polling {
-//	 delay_seconds = 30
-//	 max_attempts = 50
-// }
+//
+//	aws_polling {
+//		 delay_seconds = 30
+//		 max_attempts = 50
+//	}
+//
 // ```
 //
 // JSON example:
 // ```json
-// "aws_polling" : {
-// 	 "delay_seconds": 30,
-// 	 "max_attempts": 50
-// }
+//
+//	"aws_polling" : {
+//		 "delay_seconds": 30,
+//		 "max_attempts": 50
+//	}
+//
 // ```
 type AWSPollingConfig struct {
 	// Specifies the maximum number of attempts the waiter will check for resource state.
@@ -99,7 +103,7 @@ func (w *AWSPollingConfig) WaitUntilAMIAvailable(ctx aws.Context, conn ec2iface.
 	return err
 }
 
-func (w *AWSPollingConfig) WaitUntilInstanceRunning(ctx aws.Context, conn *ec2.EC2, instanceId string) error {
+func (w *AWSPollingConfig) WaitUntilInstanceRunning(ctx aws.Context, conn ec2iface.EC2API, instanceId string) error {
 
 	instanceInput := ec2.DescribeInstancesInput{
 		InstanceIds: []*string{&instanceId},
--- a/builder/common/step_run_source_instance.go
+++ b/builder/common/step_run_source_instance.go
@@ -294,31 +294,11 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
 
 	// Set the instance ID so that the cleanup works properly
 	s.instanceId = instanceId
-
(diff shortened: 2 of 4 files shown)


yinyic (approved):

lbajolet-hashicorp (changes requested):
Hi @sparrc,

Thanks for the fix, this looks good to me overall, however I notice that the type change in the `step_run_spot_instance.go` file for the `ec2conn` will break tests.

We don't have any test for the `step_run_source_instance`, but we do for the `step_run_spot_instance`, and it creates the connection using a EC2Mock.
Because of this, the type cast fails, and the test crashes.

In order to fix this, we should accept any `ec2iface.EC2API` for the `WaitUntilInstanceRunning` function. Fortunately it is something we define, so this should be fixable.

In its current state then we cannot merge this, but once this is fixed, I'll be glad to give this PR another review, and eventually merge it.

Do you need any assistance fixing this problem? If so do let me know, I'll do a pass on the code and fix the typing problem.

Thanks again for this fix, hopefully we can merge it soon!

sparrc:
Thanks for the review, I will take a look at that, I think I should be able to fix it

sparrc:
Updated to PR to use `ec2iface.EC2API` down into `WaitUntilInstanceRunning`, thanks again for the review @lbajolet-hashicorp !

sparrc:
Now it seems like the mocks are panicking when `ec2conn.WaitUntilInstanceRunningWithContext` is called. Not exactly sure why that is though? Do the mocks need to be regenerated?

lbajolet-hashicorp (approved):
Hi @sparrc,

Thanks for following up with this, the code in its current state looks good to me, and tests are passing, I'll merge this right away.

PR #309, opened 2023-01-14
Instance SSH access: use all attempts to get public IP

sparrc:
Previously the ssh access IP would fallback to the private ip address on the first "try" if a public IP is not available yet.

This changes the logic to only fallback to private IP address if we are on the final attempt at finding an IP address.

Precedence of always preferring public IP is left unchanged.

related to #40 and #308

### DIFF
--- a/builder/common/ssh.go
+++ b/builder/common/ssh.go
@@ -60,7 +60,9 @@ func SSHHost(e ec2Describer, sshInterface string, host string) func(multistep.St
 			} else if i.VpcId != nil && *i.VpcId != "" {
 				if i.PublicIpAddress != nil && *i.PublicIpAddress != "" {
 					host = *i.PublicIpAddress
-				} else if i.PrivateIpAddress != nil && *i.PrivateIpAddress != "" {
+				} else if i.PrivateIpAddress != nil && *i.PrivateIpAddress != "" && j == tries {
+					// if this is the final try, fallback to private ip address, otherwise continue
+					// trying to get a public ip address.
 					host = *i.PrivateIpAddress
 				}
 			} else if i.PublicDnsName != nil && *i.PublicDnsName != "" {


sparrc:
For context, https://github.com/hashicorp/packer-plugin-amazon/pull/308 is the PR that fixes the root cause of our issue, but I thought I'd submit this change as another potential PR.

I believe most users, when not specifying "ssh_interface", would expect this function to try to find a public IP first before falling back to private IP.

lbajolet-hashicorp (commented):
Hi @sparrc,

The logic sounds good to me, I'm not sure if we should merge this as the change in behaviour may be surprising to users, though I tend to think it is a sound approach.

I'll discuss this with the rest of the team, we'll come back to this PR once we settle on this subject.

Thanks for the contribution!

lbajolet-hashicorp:
Hi @sparrc,

Regarding the code we've agreed that the logic looks sound, to highlight the potential change in behaviour, we're labelling this as a bump minor, that way clients will be able to pin the current version if they want to keep the existing behaviour.

On another note, have you experienced the issue? If so, could you provide a template to reproduce the issue?
If possible we'd like to add an acceptance test for this before merging.

Thanks again!

sparrc:
Yes I did experience this issue, it was the same issue as https://github.com/hashicorp/packer/issues/10347

Now with HEAD of main I can't repro the issue (because of https://github.com/hashicorp/packer-plugin-amazon/pull/308).

But if you checkout a commit before #308 was merged, and launched an ami build with a spot instance, then you would see packer trying to SSH to the instance's private ip address almost every time, since at that point the public ip address is not ready or assigned yet.

lbajolet-hashicorp (approved):
Hi @sparrc,

Understood, it makes sense that with the previous PR being merged we add the extra waiting time for the instance to become available, it leaves more time for AWS to set a public IP for the instance.

The change makes sense even if we can't reliably reproduce it, we won't add an acceptance test, but I think the fix still holds, so we'll merge this PR, and highlight the change in the upcoming release of the plugin.

Thanks again for the fix and the explanations!

PR #312, opened 2023-01-23
Release v1.2.0

lbajolet-hashicorp:
Prepare releasing version 1.2.0 of the plugin

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -4,12 +4,12 @@ import "github.com/hashicorp/packer-plugin-sdk/version"
 
 var (
 	// Version is the main version number that is being run at the moment.
-	Version = "1.1.7"
+	Version = "1.2.0"
 
 	// VersionPrerelease is A pre-release marker for the Version. If this is ""
 	// (empty string) then it means that it is a final release. Otherwise, this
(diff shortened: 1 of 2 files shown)


nywilken (approved):
LGTM. Before releasing could you double check that all closed PRs have been labeled accordingly so that they show up properly in the generated release notes.

version/version.go
```diff
@@ -4,7 +4,7 @@ import "github.com/hashicorp/packer-plugin-sdk/version"
 
 var (
 	// Version is the main version number that is being run at the moment.
-	Version = "1.1.7"
+	Version = "1.2.1"
```
nywilken, line 7: I know that with the tag goreleaser will override the version of the binary. But if you check out this plugin @v1.2.0 and build locally the version will be 1.2.2-dev which might be confusing. Not worth changing at this moment since this is just my opinion but something to think about.

lbajolet-hashicorp: Oh yeah so, the tag will be on the commit what bumps the version to 1.2.0 and removes the `dev` suffix, I just bundled both the version update for the release and the preparation for the next one to make a single PR to release and prepare.
Would it be clearer if I split this in two different PRs?

PR #315, opened 2023-01-25
Skip uploading ssh pub key when AWS key pair is used

### LINKED ISSUE #313: PR #296 requires ec2-instance-connect:SendSSHPublicKey even if you're not specifying a private key
#### Overview of the Issue
PR #296 says that "If the user does not specify a private key, then works like the previous implementation."

We are not specifying a private key, but our Packer builds are now failing because the IAM policy of the role running Packer doesn't have `ec2-instance-connect:SendSSHPublicKey` permissions.

The only parameters we're specifying are `ssh_username="ubuntu"` and `ssh_interface="session_manager"`.

### Plugin and Packer version

Plugin version is `github.com/hashicorp/amazon v1.2.0`

Glyphack:
Skip uploading SSH public key to the instance in case the AWS key pair is used.

In case key pair is generated and it's not specified by the user uploading the key is not necessary.

Since this issue can be identified with the roles running packer itself, I'm unsure if I can reflect this change in tests. Any help is appreciated. I tested it on my local machine with two configs:

- With SSM and no privateKeyFile: Does not upload the pub key
- With SSM and with privateKeyFile: Uploads the pub key

Closes #313

### DIFF
--- a/builder/common/step_create_ssm_tunnel.go
+++ b/builder/common/step_create_ssm_tunnel.go
@@ -87,7 +87,7 @@ func (s *StepCreateSSMTunnel) Run(ctx context.Context, state multistep.StateBag)
 		}
 	}()
 
-	if len(s.SSHConfig.SSHPrivateKey) != 0 {
+	if len(s.SSHConfig.SSHPrivateKey) != 0 && s.SSHConfig.SSHKeyPairName == "" {
 		ui.Say("Uploading SSH public key to instance")
 		err := s.sendUserSSHPublicKey(instance, s.SSHConfig.SSHPrivateKey)
 		if err != nil {
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -540,6 +540,15 @@ func TestAccBuilder_EbsSessionManagerInterface(t *testing.T) {
 					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
 				}
 			}
+
+			logs, err := os.ReadFile(logfile)
+			if err != nil {
+				return fmt.Errorf("couldn't read logs from logfile %s: %s", logfile, err)
+			}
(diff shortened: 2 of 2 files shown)


lbajolet-hashicorp (approved):
Hi @Glyphack,

Thanks for the quick turnaround on this issue, much appreciated.

Regarding testing, I'll see to add an acceptance test for this problem so we don't run into it later on.

I'll probably push this on top of your branch when I have it running, unless you prefer I open another PR for this later.

Pre-approving the PR, the code looks good to me, thanks again!

lbajolet-hashicorp:
Hi @Glyphack,

This took longer than expected since I tried to wrestle with roles for a while, and in the end settled to checking in the logs that we indeed don't attempt to upload an SSH key when none is specified in the configuration file.

I can confirm that the acceptance test I adapted does fail on main since the SSH key is being pushed, and with your commit, it does not anymore.

I'll see to have another pair of eyes take a look at the tests before merging this PR, and if/when I have an approval, I will trigger the merge.

Thanks again for the fix!

Glyphack:
Thanks for your effort @lbajolet-hashicorp
Do you think it makes sense to check that specific upload message at the acceptance test?

I was thinking about this but it's would be a very flakey test(easily breaks with just changing the message) pl that's why I decided to not do it.

But anyway testing this is really hard since the method has side effects that are hard to test.

lbajolet-hashicorp:
Yeah unfortunately I attempted to use a profile which does not possess that permission as part of their role, but it took me way too long to understand that it still succeeded because the API call was using my session and not the SSM one, hence why I ultimately decided to settle on this approach here.
I agree that it is inherently fragile, but for the moment it'll do, especially considering it's breaking users' workflow, so I would like to merge this in quickly, and follow-up with a release short after (tentatively tomorrow).

We can think of a better way to handle this test later I would think.

nywilken (approved):
LGTM

PR #317, opened 2023-01-27
acceptance: use pure Go for generating temp keys

lbajolet-hashicorp:
Some acceptance tests require generating a temporary SSH key pair for ensuring they are sent to AWS and that we can connect through SSH to an instance with a user-provided key.

This was implemented through `ssh-keygen', part of the openssh suite of tools, but because the behaviour may change between environments, we opted to rewrite the keygen steps in pure go, by relying on the functions exposed by the `crypto' standard library.

### DIFF
--- a/builder/ebs/acceptance/utils.go
+++ b/builder/ebs/acceptance/utils.go
@@ -1,21 +1,32 @@
 package amazon_acc
 
 import (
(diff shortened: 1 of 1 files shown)


nywilken (approved):
LGTM.

PR #319, opened 2023-01-27
Harden EBS ssh key tests

lbajolet-hashicorp:
The acceptance test that would generate an SSH key prior to running it was failing on Windows machines in CI as it made the process timeout.

While this is not clear why that happens, we fixed the issue earlier by adding this extra check prior to running the test, so it is not run when running normal tests in CI.

Because we now generate this SSH key directly from Go code, and not by relying on ssh-keygen, we can remove this as we don't risk encountering this problem anymore.

In addition, the EBS test in which we push the public key corresponding to the private key we add in the configuration on an SSM-backed connection would not check that the SSH public key upload message was in the output, potentially leading to some inconsistencies in the future.

To avoid this from happening at some point, we add an extra check that the message was indeed in the output.


nywilken (approved):
lgtm

Issue #286, opened 2022-11-02
Add tags to the temporary IAM role and network interface


pascal-hofmann:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

Currently the temporary IAM role and network interface created by packer (the latter indirectly trough the EC2 instance) do not get any tags at all. It would be great if these would be tagged with `run_tags`.

#### Use Case(s)

- Cost reports (it should be possible to add cost tags to all recources).

PR #323, opened 2023-02-02
iam: attach tags to instance role and profile

### LINKED ISSUE #322: EBS builder - Feature to add tags to temporary role created
_This issue was originally opened by @l3LUM in https://github.com/hashicorp/packer/issues/12236 and has been migrated to this repository. The original issue description is below._

---
I want to use the [temporary_iam_instance_profile_policy_document](https://developer.hashicorp.com/packer/plugins/builders/amazon/ebs#temporary_iam_instance_profile_policy_document) on the [amazon-ebs builder](https://developer.hashicorp.com/packer/plugins/builders/amazon/ebs) to authorise instance managed by packer to get access to a bucket.

My compagny enforce all new IAM roles to have specific tags, but I couldn't find any argument to add a tag to the temporary role created. So Packer failed at start when creating temporary resources.

Is there a workaround to add tags to the temporary created role ?

Or could you make a feature that adds a "tags" field for the temporary_iam_instance_profile_policy_document option ?

Glyphack:
RunTags value is used to set tags for resources created by packer except for Iam role and profile. This PR adds the tags when creating these resources.
Resolves #322

This will also address part of #286

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -278,7 +278,7 @@ type RunConfig struct {
 	//
 	// `security_group_ids` take precedence over this.
 	SecurityGroupFilter SecurityGroupFilterOptions `mapstructure:"security_group_filter" required:"false"`
-	// Key/value pair tags to apply to the generated key-pair, security group, snapshot and instance
+	// Key/value pair tags to apply to the generated key-pair, security group, iam profile and role, snapshot and instance
 	// that is *launched* to create the EBS volumes. The resulting AMI will also inherit these tags.
 	// This is a [template
 	// engine](/packer/docs/templates/legacy_json_templates/engine), see [Build template
--- a/builder/common/step_iam_instance_profile.go
+++ b/builder/common/step_iam_instance_profile.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/hashicorp/packer-plugin-sdk/multistep"
 	packersdk "github.com/hashicorp/packer-plugin-sdk/packer"
+	"github.com/hashicorp/packer-plugin-sdk/template/interpolate"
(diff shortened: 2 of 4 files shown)


Glyphack:
I still have to write tests for this but wanted to get some initial thoughts. 
To me it seems like we can use generics on the [TagMap type](https://github.com/hashicorp/packer-plugin-amazon/pull/323/files#diff-75442cb9df8b9d9388a6bf861f5709d025953408e37a9fc168f9f95e762a9a22R47) to convert it to AWS SDK types.

lbajolet-hashicorp (approved):
Hi @Glyphack,

The code looks simple enough to me, I think we can merge it soon.

I was puzzled by the inclusion of  `interpolate` in the iam_instance_profile step, but after re-reading the code it seems on par with the other steps that include tags in the output, so it's good.

Regarding testing, I wouldn't worry too much about it, from what I see the EBS tags aren't tested as well, and the approach looks similar enough that we can skip them here, especially since there's not a lot we can do locally.
Maybe we can try writing an acceptance test to be sure it is well integrated, but that can come later I think.
If you have an idea of tests to write, please let me know, I'll hold off on merging this until you've had time to add those to the PR.

About generics, this should be supported since we're building on 1.18 at minimum for this plugin, so we can think of leveraging them for the TagMap, although I'm curious, since this is a `map[string]string`, how would you think we can use them?

Anyway all that to say: thanks again for the contribution! I'll wait until your response regarding the tests, and after that I think we can merge this.

Glyphack:
Hi @lbajolet-hashicorp,

You're right about the tests. My initial thought was that we can run a check for the building phase. Now that I read the code realized the Checks are only for the post build step. Adding a test for this requires adding a new functionality to tests which I prefer to stick with your suggestion.

lbajolet-hashicorp:
Sounds good @Glyphack, I'll merge this now and move on to review your other PRs, thanks!

PR #324, opened 2023-02-05
Fix ssm bad exit status error

### LINKED ISSUE #39: aws-ebs terminating instance exits with bad status at very end of AMI creation
_This issue was originally opened by @simonw7034 as hashicorp/packer#10278. It was migrated here as a result of the [Packer plugin split](https://github.com/hashicorp/packer/issues/8610#issuecomment-770034737). The original body of the issue is below._

<hr>

#### Overview of the Issue

I am building an AMI using the aws-ebs builder. Everything works fine up until what appears to be the cleanup stage, where the "Terminating the source AWS instance" step exits with bad status (even though it appears to finish the AMI build properly), the AMI is created and stored in our VPC but the code exits prematurely and with a non-zero rc. There may be artefacts left in AWS that aren't cleaned up properly (I haven't seen but there could be temporary security groups etc. leftover) and it makes it difficult to incorporate this into a CI build process.

The general build process is: 1) launch source AMI, 2) install some dependencies with apt 3) run a shell script to change the default boot kernel and reboot 4) wait for reboot 5) carry on with further provisioning (not included in this build template)

#### Reproduction Steps

Run`packer build -except=vagrant bf_packer.json` with an appropriate AWS VPC, user permissions and an IAM instance profile with AmazonSSMManagedInstanceCore policy and AWS authentication environment variables set.

### Packer version

Packer v1.6.5

[Packer build template](https://gist.github.com/simonw7034/a27d68b2eddf4cd6c030f19d62870793)

MacBook Pro, 64-bit

[Debug log](https://gist.github.com/simonw7034/ebabe6a06c64fda526739bfcff1d4e0a)

Glyphack:
Terminate SSM session after instance is terminated
    
When the instance is terminated the started session must be terminated otherwise the start-session command will exit with -1 status.

## Previous & New Behavior

In cases where session manager config was used, after packer finished building an AMI successfully this log was printed out:

```
==> amazon-ebs.basic-example: Terminating the source AWS instance...
2023/02/04 17:42:07 packer-plugin-amazon plugin: 2023/02/04 17:42:07 ssm: PortForwarding session "xxx-08d849476039ac655" failed: Bad exit status: -1
==> amazon-ebs.basic-example: Bad exit status: -1 <-- problematic line
==> amazon-ebs.basic-example: ssm: PortForwarding session is finished
2023/02/04 17:42:07 packer-plugin-amazon plugin: 2023/02/04 17:42:07 ssm: Terminating PortForwarding session "xxx-08d849476039ac655"
==> amazon-ebs.basic-example: ssm: PortForwarding session is finished
```

This line was printed because the session was terminated after the instance was terminated, as you can see in the log lines order.

In the new behavior the instance status is checked periodically and session is closed when instance is terminated:

```
==> amazon-ebs.basic-example: Stopping the source instance... <-- notice this is stop
    amazon-ebs.basic-example: Stopping instance
==> amazon-ebs.basic-example: Waiting for the instance to stop...
2023/02/04 22:10:33 packer-plugin-amazon plugin: 2023/02/04 22:10:33 ssm: Terminating PortForwarding session "xxx-0fc5ed4a4ace5d3e7"
    amazon-ebs.basic-example: Exiting session with sessionId: xxx-0fc5ed4a4ace5d3e7.
==> amazon-ebs.basic-example: ssm: PortForwarding session is finished
```

## Notes

I wanted to make sure we won't terminate the session at the wrong time. I checked the [docs on instance life-cycle](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html) and tried out a few scenarios to test the `describe-instance-status` behavior.

The result is that in `shutting-down`, `Terminated` and `Stopped` states (The red boxes in the doc) the API returns empty response which is what I used to identify if the session must be closed. It does not trigger a terminate when instance is restarted. Is there any case where we stop the instance but then again resume the build process? If this is present then this code is not right, and we can look into other solutions.

## Implementation

Inside the `Start` session method another coroutine is started which checks for instance status every 2 seconds.
When instance status is not present we terminate the session. in case the API call is not clear I adopted this pattern [from here](https://github.com/awslabs/amazon-ec2-instance-qualifier/blob/main/pkg/resources/instance.go#L49).

## Alternatives

Another approach would be to communicate between steps. In this method a signal can be sent to `CreateSSMTunnelStep` from `CleanUp` method of `RunInstance`. I did not proceed with this because I did not find any previous usage of this pattern I was not sure if it's an anti-pattern or not.

Closes #39


builder/common/ssm/session.go
```diff
@@ -84,21 +86,48 @@ func (s Session) getCommand(ctx context.Context) ([]string, string, error) {
...
 		if sessionID != "" {
 			defer func() {
-				_, err := s.SvcClient.TerminateSession(&ssm.TerminateSessionInput{SessionId: aws.String(sessionID)})
-				if err != nil {
-					ui.Error(fmt.Sprintf("Error terminating SSM Session %q. Please terminate the session manually: %s", sessionID, err))
+				s.terminateSession(sessionID, ui)
+			}()
+			// If the instance is terminated the session must be terminated as well
+			// Otherwise the start-session command will exit with status -1
+			go func() {
```
lbajolet-hashicorp, line 116: Seeing this is in a loop, this coroutine will be spawned multiple times, but from my understanding, it should only be invoked once, right?
The deferred function above as well I'd think, since we only kill the session when we leave the parent function.

builder/common/ssm/session.go
```diff
@@ -84,21 +86,48 @@ func (s Session) getCommand(ctx context.Context) ([]string, string, error) {
...
+				for {
+					instanceState, err := s.Ec2Conn.DescribeInstanceStatus(&ec2.DescribeInstanceStatusInput{
+						InstanceIds: []*string{aws.String(s.InstanceID)},
+					})
+					if err != nil {
+						log.Printf("ssm: Error describing instance status: %s", err)
+					}
+					// if no instance status is returned, the instance is terminated
+					if len(instanceState.InstanceStatuses) == 0 {
```
lbajolet-hashicorp, line 125: In addition to this we could also probably terminate the session when the instance is being shut down, what do you think?

builder/common/ssm/session.go
```diff
@@ -84,21 +86,48 @@ func (s Session) getCommand(ctx context.Context) ([]string, string, error) {
...
+						exitSession = true
+						s.terminateSession(sessionID, ui)
```
lbajolet-hashicorp, line 127: This should probably not be necessary, as the deferred function already closes the session when we exit `Start`

lbajolet-hashicorp (changes requested):
Hi @Glyphack,

Thanks for this PR! I left a few comments on the code for `Start`, I think we can probably re-organize the statements, especially the deferred/coroutine since they'll be registered/spawned multiple times from what I understand, while they should only be invoked once.

Feel free to address those comments when you have time, I'll do another review pass when you're ready

builder/common/ssm/session.go
Glyphack: There's one case which I'm not sure how to handle without this:

1. code enters the loop and creates the session this blocks the coroutine when it reaches the `localexec.RunAndStream(cmd, ui, nil)`.
2. the status checker finds out the instance is terminated and set `exitSession` to true.
3. the coroutine is still waiting for the `localexec.RunAndStream(cmd, ui, nil)` to exit. This actually never exits until the command fails and the Bad exit status happens.

builder/common/ssm/session.go
Glyphack: Good suggestion

builder/common/ssm/session.go
Glyphack: You're right. Does the following approach make sense?

I can create a sessionIds slice, this is used to keep record of sessions created. because it's a loop we might create more than one session. 

when the instance is in termination the session must be immediately terminated, and we cannot rely on the defer([same as this comment](https://github.com/hashicorp/packer-plugin-amazon/pull/324/files#r1103209580)). This is where we need those sessionIds to terminate all of them.

This second approach only has one coroutine and that one will terminate all the sessions created at once.

Glyphack: One more thing, if we do the termination immediately then the defer is redundant.

Glyphack:
Thanks for the review @lbajolet-hashicorp 

I understand your point here. I think we can extract the goroutine out of the loop to only spawn it once. In this case we have to keep an array of every sessionid created inside the loop to terminate them all in case of a shutdown.

This is my observations, let me know where can I improve the code:

Only setting the exitSession to true is not enough to prevent this bad exit status. Although the defer is registering a terminatation but the run command operation is stuck until it hit an error then the defer runs.
It would be good if we could cancel the run command operation but since we can't we have to call the terminate session when instance shuts down.

lbajolet-hashicorp:
Hi @Glyphack,

Sorry for not getting back to you sooner, I'm working on something unrelated for the moment so I didn't have time to come back to you regarding this PR, but I will definitely do another pass this week.

builder/common/ssm/session.go
lbajolet-hashicorp: Regarding this thread, I misunderstood the intent here, my bad, I thought this was only one session ID, but from the looks of it, there are indeed multiple sessions created, one per loop.

In this case, I don't think we'll benefit from storing the session IDs and purging them after we're done, so I think we can continue with the proposed approach.

builder/common/ssm/session.go
lbajolet-hashicorp: I might be wrong, but if the `DescribeInstanceStatus` call fails, `instanceState` may be `nil`, right? If so, this will crash the plugin as deferencing a `nil` object causes a NRE, and since it's in a coroutine and it's never caught, the whole plugin will come crashing down if/when that happens.

We should check for that here.

lbajolet-hashicorp: Also, in terms of permissions, I think it'll be rare (maybe impossible in our context), but if someone cannot perform the `DescribeInstanceStatus` call, this coroutine will never exit, will it?
This will also be true if for some reason the API call always ends with an error, as such, I would suggest we add a way to send a message from the outside, for example when `localexec.RunAndStream` exits naturally, this way we can stop checking periodically, and exit the coroutine when that happens.

lbajolet-hashicorp: If we are waiting on outside events as well, I suggest we use a `time.Ticker` for the periodic checks. This way we can use a `select` statement, and can react immediately to whichever event triggers on a channel, and immediately quit if we do receive that message.

builder/common/ssm/session.go
lbajolet-hashicorp: I also agree with your sentiment, since we terminate the session here, we can remove the deferred function entierly.

builder/common/ssm/session.go
lbajolet-hashicorp: If we remove the `defer` as pointed out above, we can indeed terminate here, no objection on my part!

lbajolet-hashicorp (commented):
Hi @Glyphack,

Sorry again for the long wait, I appreciate your patience on this subject.

I continued the discussion on some of the threads, I misunderstood the way the function worked regarding the `sessionID`, so I think we can continue iterating on the current approach.

Please take a look at my comments, and let me know if you find my suggestions relevant.
If you need a hand in implementing this, let me know and I'll be happy to take a jab at this.

Thanks again!

builder/common/ssm/session.go
Glyphack: Good point, thanks. I'm going to document the permission as well in the docs. In case someone gets that error again they can check if they have the permission.

Glyphack:
Hi @lbajolet-hashicorp

I applied your comments, and learned a lot in figuring out how to do the for-select timer in a non blocking way. Thanks for the tip.

I have tested the final changes with a simple build file, and seems be working as expected.

docs-partials/builders/aws-session-manager.mdx
```diff
@@ -100,3 +101,11 @@ In order for Packer to start and end sessions that connect you to your managed i
 #### IAM instance profile for Systems Manager
 
 By default Systems Manager doesn't have permission to perform actions on created instances so SSM access must be granted by creating an instance profile with the `AmazonSSMManagedInstanceCore` policy. The instance profile can then be attached to any instance you wish to manage via the session-manager-plugin. See [Adding System Manager instance profile](https://docs.aws.amazon.com/systems-manager/latest/userguide/setup-instance-profile.html#instance-profile-add-permissions) for details on creating the required instance profile.
+
+#### Permissions for Closing the Tunnel
+
+To close the SSM tunnels created, this plugin relies on being able to call
+[DescribeInstanceStatus](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstanceStatus.html).
+In case this is not possible you will might get `Bad exit status: -1` message in the logs.
```
lbajolet-hashicorp, line 109:
```suggestion
In case this is not possible you might see a `Bad exit status` message in the logs.
```

builder/ebs/builder_acc_test.go
```diff
@@ -549,6 +549,10 @@ func TestAccBuilder_EbsSessionManagerInterface(t *testing.T) {
 				return fmt.Errorf("SSH key was uploaded, but shouldn't have been")
 			}
 
+			if strings.Contains(string(logs), "Bad exit status: -1") {
```
lbajolet-hashicorp, line 552: Just thought about the exit code here, I wonder if it'll always be `-1` and if it may not change in the future for some reason. I think we can be slightly more resilient by only checking for `Bad exit status` in the output here.

lbajolet-hashicorp (approved):
Thanks for the quick reroll on this PR @Glyphack, much appreciated!

The current approach looks good to me, I left a few last comments on the code, feel free to address them, and we can finalise this PR.

I'm pre-approving it in the current state for a later merge.

lbajolet-hashicorp:
Thanks for the last updates @Glyphack,

Merging this code now.

PR #326, opened 2023-02-08
Add run_tags to ec2 instance network interfaces

### LINKED ISSUE #286: Add tags to the temporary IAM role and network interface

Glyphack:
This PR adds run_tags to ec2 instance network interfaces.

From the [RunInstanceInput](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#RunInstancesInput) we can see that the NetworkInterface tags are also applied from the TagSpecification field.

To test the code block in the IsRestricted, I manually turned it on and built an AMI.

Closes #286

### DIFF
--- a/builder/common/step_run_source_instance.go
+++ b/builder/common/step_run_source_instance.go
@@ -386,13 +386,13 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
 
 		if len(ec2Tags) > 0 {
 			for _, networkInterface := range instance.NetworkInterfaces {
+				log.Printf("Tagging network interface %s", *networkInterface.NetworkInterfaceId)
 				_, err := ec2conn.CreateTags(&ec2.CreateTagsInput{
 					Tags:      ec2Tags,
 					Resources: []*string{networkInterface.NetworkInterfaceId},
 				})
-				log.Printf("Tagging network interface %s", *networkInterface.NetworkInterfaceId)
 				if err != nil {
-					ui.Say(fmt.Sprintf("Error tagging source instance's network interface %q: %s", *networkInterface.NetworkInterfaceId, err))
+					ui.Error(fmt.Sprintf("Error tagging source instance's network interface %q: %s", *networkInterface.NetworkInterfaceId, err))
 				}
 			}
 		}
--- a/builder/common/step_run_source_instance.go
+++ b/builder/common/step_run_source_instance.go
@@ -392,7 +392,7 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
 				})
 				log.Printf("Tagging network interface %s", *networkInterface.NetworkInterfaceId)
 				if err != nil {
-					ui.Say(fmt.Sprintf("Error tagging source instance's network interface: %s, skip tagging network interfaces", err))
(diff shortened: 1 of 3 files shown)


builder/common/step_run_source_instance.go
```diff
@@ -377,6 +384,18 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
...
 
+		if len(ec2Tags) > 0 {
+			for _, networkInterface := range instance.NetworkInterfaces {
+				_, err := ec2conn.CreateTags(&ec2.CreateTagsInput{
+					Tags:      ec2Tags,
+					Resources: []*string{networkInterface.NetworkInterfaceId},
+				})
+				log.Printf("Tagging network interface %s", *networkInterface.NetworkInterfaceId)
+				if err != nil {
+					ui.Say(fmt.Sprintf("Error tagging source instance's network interface: %s, skip tagging network interfaces", err))
```
Glyphack, line 395: I think this should be a silent error to not make any breaking changes here. So we can print the error and get some feedback if someone got this error at all.

lbajolet-hashicorp: I don't think this qualifies as silent error, but I agree that we shouldn't halt the build process if that happens for whatever reason.
I would maybe suggest we use `ui.Error` rather than `ui.Say` for reporting this to the user though

lbajolet-hashicorp: About the message though, I would maybe suggest rephrasing it a little bit so users know which interface failed to be tagged
```suggestion
					ui.Say(fmt.Sprintf("Error tagging source instance's network interface %q: %s", *networkInterface.NetworkInterfaceId, err))
```

builder/common/step_run_source_instance.go
```diff
@@ -377,6 +384,18 @@ func (s *StepRunSourceInstance) Run(ctx context.Context, state multistep.StateBa
 			return multistep.ActionHalt
 		}
 
+		if len(ec2Tags) > 0 {
+			for _, networkInterface := range instance.NetworkInterfaces {
+				_, err := ec2conn.CreateTags(&ec2.CreateTagsInput{
+					Tags:      ec2Tags,
+					Resources: []*string{networkInterface.NetworkInterfaceId},
+				})
+				log.Printf("Tagging network interface %s", *networkInterface.NetworkInterfaceId)
```
lbajolet-hashicorp, line 393: Maybe this should be moved before the creation of the tags?

lbajolet-hashicorp (approved):
LGTM @Glyphack, I left a few comments on the code, feel free to address those when you can and I'll do a second pass.

Regarding testing, I presume those tags for the network instance won't be part of the resulting AMI? That rules out acceptance testing with the current infra we have if so, so that's unfortunate :/

Glyphack:
Thanks for the review! addressed the comments.

Yeah the tags can only be checked in the build step(same as the role tags) and after build is finished network interface is removed.

lbajolet-hashicorp:
Thanks for the speedy reroll @Glyphack.

The code looks good to me as it stands, because of the merge from `main` into this PR, I won't be able to rebase/merge, so for this once I will squash the changes in one commit and merge that.

Thanks again for the involvement, much appreciated!

PR #327, opened 2023-02-10
Copy tags on copy ami

### LINKED ISSUE #295: Support AMI Tag Copying
Please search the existing issues for relevant feature requests, and use the
reaction feature
(https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/)
to add upvotes to pre-existing requests.

#### Community Note

#### Description

[AWS just announced](https://aws.amazon.com/about-aws/whats-new/2022/11/amazon-ec2-enables-customer-created-tags-included-during-image-copy/) support for including tags when copying AMIs. This is supported by a new [CopyImageTags](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CopyImage.html#API_CopyImage_RequestParameters) boolean in the ec2.CopyImage action.

It would be useful to expose this in Packer's AMI configuration settings.

#### Use Case(s)

Using `ami_regions` to copy the resulting AMI to other regions leaves those copies without tags. Tags can be useful for downstream processes to identify the intended AMI to use, for example with the `aws_ami` Terraform datasource's filtering. Historically, tagging AMI copies in other regions required an external process. AWS now supports this with the new boolean, so users will no longer need to tag the AMIs themselves. 

#### Potential configuration

```hcl
copy_image_tags = true
```

#### Potential References

https://github.com/aws/aws-sdk-go/compare/v1.44.141...v1.44.142#diff-1082fdae378a5abc33332afae3c56453249b51a8309b4e3d15278e8c23471d2fR61657

lbajolet-hashicorp:
As pointed out in #295, the tags specified as `run_tags` in a template are applied to the resulting base AMI, but are not copied to the AMIs that are copied from it to other regions.

This does not apply to tags specified through the `tags` attribute since they are manually copied through an extra step, after the AMIs are created and copied.

To fix the first use case, Amazon added support for copying the tags when an AMI is copied to another region, so we leverage that in this PR.

Closes #295

### DIFF
--- a/builder/common/step_ami_region_copy.go
+++ b/builder/common/step_ami_region_copy.go
@@ -202,12 +202,14 @@ func (s *StepAMIRegionCopy) amiRegionCopy(ctx context.Context, state multistep.S
 	if err != nil {
 		return "", snapshotIds, err
 	}
+	t := true
 	resp, err := regionconn.CopyImage(&ec2.CopyImageInput{
 		SourceRegion:  &source,
 		SourceImageId: &imageId,
 		Name:          &name,
 		Encrypted:     encrypt,
 		KmsKeyId:      aws.String(keyId),
+		CopyImageTags: &t,
 	})
 
 	if err != nil {
--- a/builder/common/step_ami_region_copy_test.go
+++ b/builder/common/step_ami_region_copy_test.go
@@ -32,6 +32,9 @@ type mockEC2Conn struct {
 }
 
 func (m *mockEC2Conn) CopyImage(copyInput *ec2.CopyImageInput) (*ec2.CopyImageOutput, error) {
+	if !*copyInput.CopyImageTags {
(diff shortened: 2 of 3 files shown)


builder/common/step_ami_region_copy.go
```diff
@@ -202,12 +202,14 @@ func (s *StepAMIRegionCopy) amiRegionCopy(ctx context.Context, state multistep.S
 	if err != nil {
 		return "", snapshotIds, err
 	}
+	t := true
```
JenGoldstrich, line 205: Might be worth updating step_ami_region_copy_test for this to make sure true is passed in in the units too

JenGoldstrich (approved):
feel free to add the suggested unit test or not if you disagree and think the acceptance test is enough

builder/common/step_ami_region_copy.go
lbajolet-hashicorp: Good call! I've added this to the tests, now we always check if it is properly set (and from the looks of it, it is!).

PR #337, opened 2023-02-27
docs: add other tagged resources to run_tags description

Glyphack:
I realized after updating the run_config documents in #326 & #323 I forgot to run the packer-sdc. I just learned what does it do.


lbajolet-hashicorp (approved):
LGTM, thanks for the fix @Glyphack !

Issue #314, opened 2023-01-25
Add configuration to enable NitroTPM support when registering an AMI


alex-torok:
#### Description

[NitroTPM](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/nitrotpm.html) support allows an AMI to be used with instances that have a virtual TPM module for securely storing secrets and providing secure boot. According to AWS docs, TPM support must be enabled as an option when the AMI is registered.

#### Use Case(s)

Supporting using TPM on EC2.

#### Potential configuration

```
source "amazon-ebs" "my_ami" {
  ...
  tpm_support = "v2.0"
  ...
}
```

#### Potential References

* AWS docs about [Registering an AMI with TPM support through the aws CLI](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-support-on-ami.html)
* I'm guessing the change will need to happen when the AMI is registered [here](https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/chroot/step_register_ami.go). I was looking for a way to set this for the ebs builder, but it looks like that one does not perform the RegisterImage step. I'm not familiar with the intricate details of how building/registering an AMI works, so perhaps it isn't possible to set this when building an AMI from another AMI?
* I believe that the option that will need to be set is `RegisterImageInput.TpmSupport = "v2.0"`

ccravens:
Great feature suggestion, I think I'm gonna give this a shot at this fork here: https://github.com/ossys/packer-plugin-amazon

ccravens:
PR Submitted: https://github.com/hashicorp/packer-plugin-amazon/pull/339

lbajolet-hashicorp:
Hi @glachac-safelishare,

Good call, the issue was automatically closed when we merged the PR for the other builders, I agree though that it would be good for EBS to support it eventually. Right now we're stuck a bit because of how we create AMIs using this builder, since we're using `CreateImage` instead of snapshotting + `RegisterImage`.

For similar problems in the past, we've had Amazon add the capability to other API calls, which we can then leverage for EBS. Alternatively, we could also look into supporting this workflow for the builder as well, but as I mentioned in the PR, I would tread carefully with this approach since it would be easy to break builds when we do.

All in all, I'll reopen this issue so we can keep an eye on it for EBS, and we can decide the approach later.

Thanks for pinging us on this!

eamonnlambda:
Hi, all.

[...]

PR #339, opened 2023-03-06
Tpm support

ccravens:
Adding AWS NitroTPM support to AMI. More information can be found here:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-support-on-ami.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html

There is a flag that is not currently supported by the Packer AMI Builder: `--tpm-support v2.0`

This update is meant to resolve that. I am not a packer plugin developer, but added what I believe to be the basic necessary code and tried my best to maintain the compatibility of the code based on the other options that are currently there. I want to see this feature as soon as possible so if there are updates that need to be made please let me know.

Both `go build` and `make dev` seem to work.

Testing results in the following output:
```bash
$ make test
go: downloading github.com/stretchr/testify v1.7.0
go: downloading gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
?   	github.com/hashicorp/packer-plugin-amazon	[no test files]
ok  	github.com/hashicorp/packer-plugin-amazon/builder/chroot	1.781s
ok  	github.com/hashicorp/packer-plugin-amazon/builder/common	0.520s
?   	github.com/hashicorp/packer-plugin-amazon/builder/common/awserrors	[no test files]
?   	github.com/hashicorp/packer-plugin-amazon/builder/common/ssm	[no test files]
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebs	5.617s
?   	github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance	[no test files]
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebssurrogate	2.065s
ok  	github.com/hashicorp/packer-plugin-amazon/builder/ebsvolume	1.657s
ok  	github.com/hashicorp/packer-plugin-amazon/builder/instance	2.584s
ok  	github.com/hashicorp/packer-plugin-amazon/datasource/ami	1.382s
ok  	github.com/hashicorp/packer-plugin-amazon/datasource/parameterstore	1.011s
ok  	github.com/hashicorp/packer-plugin-amazon/datasource/secretsmanager	0.606s
?   	github.com/hashicorp/packer-plugin-amazon/post-processor/import	[no test files]
?   	github.com/hashicorp/packer-plugin-amazon/version	[no test files]
```

Closes https://github.com/hashicorp/packer-plugin-amazon/issues/314

### DIFF
--- a/builder/common/block_device.go
+++ b/builder/common/block_device.go
@@ -33,21 +33,25 @@ const (
 // HCL2 example:
 //
 // ```hcl
-// launch_block_device_mappings {
-//     device_name = "/dev/sda1"
-//     encrypted = true
-//     kms_key_id = "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
-// }
+//
+//	launch_block_device_mappings {
+//	    device_name = "/dev/sda1"
+//	    encrypted = true
+//	    kms_key_id = "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
+//	}
+//
 // ```
 //
 // JSON example:
 // ```json
 // "launch_block_device_mappings": [
-//   {
-//      "device_name": "/dev/sda1",
-//      "encrypted": true,
-//      "kms_key_id": "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
-//   }
+//
+//	{
+//	   "device_name": "/dev/sda1",
+//	   "encrypted": true,
+//	   "kms_key_id": "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
+//	}
+//
 // ]
 // ```
 //
@@ -56,7 +60,6 @@ const (
 //
 // Documentation for Block Devices Mappings can be found here:
 // https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/block-device-mapping-concepts.html
-//
 type BlockDevice struct {
 	// Indicates whether the EBS volume is deleted on instance termination.
 	// Default false. NOTE: If this value is not explicitly set to true and
--- a/docs-partials/builder/chroot/Config-not-required.mdx
+++ b/docs-partials/builder/chroot/Config-not-required.mdx
@@ -159,6 +159,10 @@
   more information. Defaults to `legacy-bios` when `ami_architecture` is `x86_64` and
   `uefi` when `ami_architecture` is `arm64`.
 
+- `tpm_support` (string) - The TPM Support mode. Valid options are `v2.0`. See the documentation on
+  [NitroTPM Support](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-support-on-ami.html) for
+  more information. Defaults to `v2.0`.
+
 - `uefi_data` (string) - Base64 representation of the non-volatile UEFI variable store. For more information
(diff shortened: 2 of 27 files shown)


docs-partials/builder/chroot/Config-not-required.mdx
```diff
@@ -159,6 +159,10 @@
   more information. Defaults to `legacy-bios` when `ami_architecture` is `x86_64` and
   `uefi` when `ami_architecture` is `arm64`.
 
+- `tpm_support` (string) - The TPM Support mode. Valid options are `v2.0`. See the documentation on
```
lbajolet-hashicorp, line 162: The doc snippet looks good to me, for completeness I'd suggest adding this to all the builders that support the feature, so in addition to the `chroot` builder, the `ebssurrogate` and the `instance` builder docs should add this to their sections.

post-processor/import/post-processor.go
```diff
@@ -50,6 +50,7 @@ type Config struct {
 	Format          string            `mapstructure:"format"`
 	Architecture    string            `mapstructure:"architecture"`
 	BootMode        string            `mapstructure:"boot_mode"`
+	TpmSupport      string            `mapstructure:"tpm_support"`
```
lbajolet-hashicorp, line 53: This is added to the post-processor here, but I don't see the related step being updated, is this voluntary?

builder/common/block_device.go
```diff
@@ -33,21 +33,25 @@ const (
...
-//     encrypted = true
-//     kms_key_id = "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
-// }
+//
+//	launch_block_device_mappings {
+//	    device_name = "/dev/sda1"
+//	    encrypted = true
+//	    kms_key_id = "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
+//	}
+//
```
lbajolet-hashicorp, line 42: While I generally am not against using a linter for cleaning-up files, in this case, updating the documentation feels out of the scope of this PR, and makes the diff harder to read.
Could you please leave the doc changes out of this PR? If you think they are relevant, maybe open a separate PR for us to review, so this doesn't become a blocker for this one?

lbajolet-hashicorp (approved):
Hi @ccravens,

Thanks for the PR! The code overall looks good to me, I left a couple of nitpicks on the docs and the changes to the code documentation, while linting is appreciated, in this case since it is intertwined with the proposed enhancement, I think this is detrimental to the comprehension of the change, and therefore suggest that this is moved to a separate PR so we can independently take a look at it.

Aside from that, I see that the `ebs` builder is left out of this PR, I presume since it does not use the `RegisterAMI` API call to finalise the AMI creation. This is not the first time a feature is left out for this specific builder unfortunately, I think we should track this work independently, and refactor the logic for all builders so they're on an equal footing regarding this method to create AMIs.

If anything regarding testing, while this won't work locally completely, I would suggest adding an acceptance test to ensure the feature works as expected. We don't have anything for builders like chroot or instance, but we do have an acceptance test structure for `ebssurrogate`.
Since this involves creating resources (and using IAM roles that are permissive in the environment tests are run), I understand if this is not something you want to do, in this case  let me know, and I will take some time to do this in the coming days.

Pre-approving the PR for a later merge, I will circle back on it when you have addressed my comments.

Thanks again for the contribution!

lbajolet-hashicorp:
Superseded by #379

PR #346, opened 2023-03-21
Imds support ebs

### LINKED ISSUE #310: Support IMDSv2 configuration for the `ebs` builder
#### Community Note

#### Description

This is a follow-up to https://github.com/hashicorp/packer-plugin-amazon/issues/277, which added this feature to the `ebssurrogate`, `instance` and `chroot` builders. Unfortunately there doesn't seem to be a trivial way to add it to the `ebs` builder, because the underlying API used is different (`RegisterImage` vs `CreateImage`):

> On another note, I understand this IMDSv2 setting is available on the `RegisterImage` API call, hence why `ebssurrogate`, `instance` and `chroot` are the only ones updated, but I wonder if this will impact those that build with `ebs` or `ebsvolume`?

(from the PR closing #277: https://github.com/hashicorp/packer-plugin-amazon/pull/278#issuecomment-1318912104)

#### Use Case(s)

Same as with #277: Improve security.

#### Potential configuration

Should use the same syntax/elements as #278 added for the other builders.

#### Potential References

See #277 and #278.

lbajolet-hashicorp:
Recently AWS started supporting setting the ImdsSupport attribute on all AMIs after their creation, and not only when registering it.

We leverage this change in this PR to enable the `imds_support` attribute on all builders that create AMIs, including EBS.

Closes #310

### DIFF
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -202,11 +202,6 @@ type Config struct {
 	// Base64 representation of the non-volatile UEFI variable store. For more information
 	// see [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/uefi-secure-boot-optionB.html).
 	UefiData string `mapstructure:"uefi_data" required:"false"`
-	// Enforce version of the Instance Metadata Service on the built AMI.
-	// Valid options are unset (legacy) and `v2.0`. See the documentation on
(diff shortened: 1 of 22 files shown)


JenGoldstrich (approved):
Nice, chatted a bit offline to understand this feature

robin-fordham:
Awesome work! really appreciate everyone's efforts!

Does anyone know when the next release will be, so we can start consuming this new feature?

PR #347, opened 2023-03-24
ebs: add support for fast-launch

### LINKED ISSUE #178: Add support for enable-fast-launch (Windows AMIs)
#### Community Note

#### Description

Amazon have recently announced [the ability to configure AMIs for faster launching](https://aws.amazon.com/about-aws/whats-new/2022/01/aws-speed-optimizations-windows-instances-ec2/). 

The CLI has been updated with [a new command](https://docs.aws.amazon.com/cli/latest/reference/ec2/enable-fast-launch.html) that enables this feature.

#### Use Case(s)

We use Packer to build our CI agent images; we'd like to avoid having to run separate steps as part of our build to configure these settings for each AMI built by Packer

#### Potential configuration

```
enable_fast_launch {
  "max-parallel-launches": 20
}
```

#### Potential References
https://aws.amazon.com/about-aws/whats-new/2022/01/aws-speed-optimizations-windows-instances-ec2/
https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/windows-ami-version-history.html#win-ami-config-fast-launch
https://docs.aws.amazon.com/cli/latest/reference/ec2/enable-fast-launch.html

lbajolet-hashicorp:
The fast-launch support for Windows AMIs is a feature that AWS added in January 2022, which pre-provisions snapshots with a clean Windows image that has performed the initial setup phase.

This means that once the snapshots are ready to use, the resulting instances that will be spawned from that AMI in the same account will be much faster to start.

This commit adds the structure required for supporting this feature for the EBS builder.

Closes #178


JenGoldstrich (approved):
chatted a bit in slack, I wish there was a way to tell OS type so we could specifically block users setting fastlaunch on Linux VMs, might be worth checking how this behaves but otherwise I think this LGTM!

builder/ebs/builder_acc_test.go
```diff
@@ -1231,6 +1323,61 @@ build {
...
+	source_ami           = "ami-00b2c40b15619f518" # Windows server 2016 base x86_64
+	instance_type        = "m3.medium"
+	region               = "us-east-1"
+	communicator         = "winrm"
+	winrm_username       = "Administrator"
+	winrm_password       = "e4sypa55!"
+	user_data_file       = "test-fixtures/ps_enable.ps"
+	fast_launch {
+		target_resource_count   = 1
+		template_id = "lt-0c82d8943c032fc0b"
```
nywilken, line 1365: Where does this template id come from?

lbajolet-hashicorp: This is the default template created by AWS when requesting fast-launch to be enabled. Since we run those acceptance tests on our dev account, this is already created and we can reuse it.
Alternatively we could use the name: `EC2FastLaunchDefaultResourceCreation-30886dd2-77ed-4e79-b9d8-f543dfa1c800`

builder/ebs/builder_test.go
```diff
@@ -242,3 +242,113 @@ func TestBuilerPrepare_IMDSSupport(t *testing.T) {
...
+					"template_name":         "name",
+					"template_version":      2,
+					"max_parallel_launches": 10,
+					"target_resource_count": 20,
+				},
+			},
+			false,
+		},
+		{
+			"Error - max parallel launches < 6",
```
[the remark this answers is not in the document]
lbajolet-hashicorp: It is according to the API docs:
> Value must be 6 or greater.

SEE: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_EnableFastLaunch.html

builder/ebs/fast_launch_setup.go
```diff
@@ -0,0 +1,106 @@
...
+	// as specified in the AWS docs.
+	LaunchTemplateName string `mapstructure:"template_name"`
+	// The version of the launch template to use
+	//
+	// If unspecified, and a template is referenced, this will default to
+	// the latest version available for the template.
+	LaunchTemplateVersion int `mapstructure:"template_version"`
+	// Maximum number of instances to launch for creating pre-provisioned snapshots
+	//
+	// If specified, must be a minimum of `6`
```
nywilken, line 37: I see this answer my question from above. Is there a default value if not specified?

builder/ebs/step_enable_fast_launch.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	ec2conn := state.Get("ec2").(*ec2.EC2)
+	ui := state.Get("ui").(packersdk.Ui)
+	amis := state.Get("amis").(map[string]string)
+
+	if s.AMISkipCreateImage {
+		ui.Say("Skipping fast-launch setup...")
+		return multistep.ActionContinue
+	}
```
nywilken, line 37: I believe this check has to happen before line 32, as "amis" is only added to state if a user does not specify the skip create image configuration. 

Alternatively, you could use the state.GetOk function to check if there is an amis map in State.

builder/ebs/fast_launch_setup.go
```diff
@@ -0,0 +1,106 @@
...
+
+import "fmt"
+
+// FastLaunchConfig is the configuration for setting up fast-launch for Windows AMIs
+//
+// NOTE: requires the Windows image to be sysprep'd to enable fast-launch. See the
+// AWS docs for more information:
+// https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/win-ami-config-fast-launch.html
+type FastLaunchConfig struct {
+	// Configure AMI with fast-launch on Windows AMIs
```
nywilken, line 14: nit: The double use of AMI tripped me up a little. What do you think of this change? 
```suggestion
	// Configure fast-launch for Windows AMIs
```

nywilken (approved):
This looks good to me. I left a few nits/questions and one possible change. But otherwise good.

builder/ebs/step_enable_fast_launch.go
lbajolet-hashicorp: You are completely right, good catch!

builder/ebs/fast_launch_setup.go
lbajolet-hashicorp: The rework sounds good to me, good idea removing tautologies like these.

PR #358, opened 2023-04-11
github: fix pathing issues for building binaries

lbajolet-hashicorp:
Running the actions to upload binaries fails with a pathing issue since relative pathing is unsupported for the upload-artifact action.


nywilken (approved):
LGTM

PR #359, opened 2023-04-11
workflows: publish all the pkg files separately

lbajolet-hashicorp:
The build_plugin_binaries workflow was pushing all the packages as one zip containing all the built binaries. To avoid this, we change the upload pattern so they all end-up as separate zip files instead in the actions, and skip the last aggregation/upload step.


nywilken (approved):
LGTM

PR #360, opened 2023-04-11
actions: upload artifact with a name

lbajolet-hashicorp:
To avoid artifacts being uploaded to a single zip, we add a name in addition to the path from which to pick-up the file.


nywilken (approved):
🚢

PR #372, opened 2023-04-21
config: add docs on permissions for public IP

lbajolet-hashicorp:
When the `associate_public_ip_address' configuration attribute is set in a config without a VPC or a Subnet, we now attempt to infer those from the default VPC and subnets.

These actions require some extra permissions, so we reflect this in the docs for the attribute.


nywilken (changes requested):
I expect to see the generated partials updated for this change. But it looks like the check is not completing the call to `make generate` before completing. Feel free to run locally and commit the changes.

lbajolet-hashicorp:
Good call regarding the partials. When re-compiling them locally I noticed the fast-launch ones were not up-to-date as well, presumably I amended them and forgot to run `make generate`.
I'll look into the workflow, it should report such discrepancies.

nywilken:
> I think we can wait until we've released 1.2.5 if we're to do it quickly.
> However, I'll add some notes to the 1.2.3 and 1.2.4 releases, unless you have an objection against that.

sounds good. looking at the other PR now. Adding to the release notes is a good call as well.

lbajolet-hashicorp:
Good, thanks for the confirmation! I'll add a paragraph to both notes right now.

PR #374, opened 2023-04-27
Fix docs for fast launch

lbajolet-hashicorp:
The fast-launch configuration block was added a couple of versions ago, but its documentation was never added to the docs page for the plugin.

This PR fixes that problem.


nywilken (approved):
LGTM

PR #375, opened 2023-05-01
docs: add public ip paragraph in IAM section

lbajolet-hashicorp:
Since the `associate_public_ip_address' attribute may need to use the `DescribeVpcs' permission in order to get the default VPC information, we add this explanation to the IAM section in the docs.


nywilken (approved):
Nice 🚢

PR #379, opened 2023-05-11
Add tpm support

### LINKED ISSUE #314: Add configuration to enable NitroTPM support when registering an AMI

alemuro:
Hello! 👋 

I'm interested in merging the PR #339 but it seems there is no activity since March. For that reason I opened a this PR with changes (and history commits) from the other PR + addressing the comments from that PR and rebasing the branch against the `main` branch. 

The author says on its MR:

> Adding AWS NitroTPM support to AMI. More information can be found here:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-support-on-ami.html
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-prerequisites.html
> 
> There is a flag that is not currently supported by the Packer AMI Builder: --tpm-support v2.0
> 
> This update is meant to resolve that. 

I'm also interested in adding this feature to the `ebs` builder. I'm not sure if this is a good opportunity to change the `ebs` builder to use the Register API, or if I should raise another PR for that use case. What do you guys think?

## Tests

I've run `go build` and `make test` and seems to work with no issues. I also tried `make testacc`

```shell
$ make test
?       github.com/hashicorp/packer-plugin-amazon       [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/chroot        2.006s
ok      github.com/hashicorp/packer-plugin-amazon/builder/common        0.933s
?       github.com/hashicorp/packer-plugin-amazon/builder/common/awserrors      [no test files]
?       github.com/hashicorp/packer-plugin-amazon/builder/common/ssm    [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebs   8.888s
?       github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance        [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebssurrogate  3.101s
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebsvolume     2.366s
ok      github.com/hashicorp/packer-plugin-amazon/builder/instance      2.629s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/ami        0.908s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/parameterstore     1.496s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/secretsmanager     1.784s
?       github.com/hashicorp/packer-plugin-amazon/post-processor/import [no test files]
?       github.com/hashicorp/packer-plugin-amazon/version       [no test files]
```

Closes #314

### DIFF
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -381,7 +381,7 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 	}
 
 	if b.config.TpmSupport != "" && b.config.TpmSupport != ec2.TpmSupportValuesV20 {
-		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(`The only valid tpm_support values are %q or the empty string`, ec2.TpmSupportValuesV20))
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(`The only valid tpm_support value is %q`, ec2.TpmSupportValuesV20))
 	}
 
 	if b.config.BootMode != "" {
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -205,6 +205,18 @@ type Config struct {
 	// Base64 representation of the non-volatile UEFI variable store. For more information
 	// see [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/uefi-secure-boot-optionB.html).
 	UefiData string `mapstructure:"uefi_data" required:"false"`
+<<<<<<< HEAD
+=======
+	// Enforce version of the Instance Metadata Service on the built AMI.
+	// Valid options are unset (legacy) and `v2.0`. See the documentation on
+	// [IMDS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html)
+	// for more information. Defaults to legacy.
+	IMDSSupport string `mapstructure:"imds_support" required:"false"`
+	// NitroTPM Support. Valid options are `v2.0`. See the documentation on
+	// [NitroTPM Support](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enable-nitrotpm-support-on-ami.html) for
+	// more information. Only enabled if a valid option is provided, otherwise ignored.
+	TpmSupport string `mapstructure:"tpm_support" required:"false"`
+>>>>>>> 5bc11e8a (Added TPM as configuration option)
 
 	ctx interpolate.Context
 }
@@ -376,6 +388,17 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 		errs = packersdk.MultiErrorAppend(errs, errors.New(`The only valid ami_architecture values are "arm64", "i386", "x86_64", or "x86_64_mac"`))
 	}
 
+<<<<<<< HEAD
+=======
+	if b.config.IMDSSupport != "" && b.config.IMDSSupport != ec2.ImdsSupportValuesV20 {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(`The only valid imds_support values are %q or the empty string`, ec2.ImdsSupportValuesV20))
+	}
+
+	if b.config.TpmSupport != "" && b.config.TpmSupport != ec2.TpmSupportValuesV20 {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(`The only valid tpm_support values are %q or the empty string`, ec2.TpmSupportValuesV20))
+	}
+
+>>>>>>> 5bc11e8a (Added TPM as configuration option)
 	if b.config.BootMode != "" {
 		err := awscommon.IsValidBootMode(b.config.BootMode)
 		if err != nil {
@@ -512,6 +535,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			PollingConfig:            b.config.PollingConfig,
 			BootMode:                 b.config.BootMode,
 			UefiData:                 b.config.UefiData,
+			TpmSupport:				  b.config.TpmSupport,
 		},
 		&awscommon.StepAMIRegionCopy{
 			AccessConfig:      &b.config.AccessConfig,
--- a/builder/chroot/builder.go
+++ b/builder/chroot/builder.go
@@ -205,18 +205,10 @@ type Config struct {
 	// Base64 representation of the non-volatile UEFI variable store. For more information
 	// see [AWS documentation](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/uefi-secure-boot-optionB.html).
 	UefiData string `mapstructure:"uefi_data" required:"false"`
-<<<<<<< HEAD
-=======
-	// Enforce version of the Instance Metadata Service on the built AMI.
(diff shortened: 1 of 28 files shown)


lbajolet-hashicorp:
Hey @alemuro,

Regarding the `ebs` builder, we'd need to change how the AMI is created, so while I'm not against the idea of providing an alternative way to do this rather than using `CreateImage`, I think it should be optional. From my understanding the only way we could do this is to first snapshot the root volume, and use `RegisterImage` to make an AMI from this snapshot.

Adding this to the builder would indeed make it possible to use this workflow and benefit from the extra options that `RegisterImage` provides, but if possible I would like to wait until we make a move on this front, at least until we have a way to run our acceptance tests reliably from a Github action, that way we can be more confident this won't break existing configs.

Alternatively, and this is more of a gamble, we could wait until there's another way to set this attribute on existing AMIs, much like what we experienced with the `IMDSv2` enforcement earlier this year. This would be preferable in my opinion due to the lesser risk to the builder, which is our most popular for this plugin, hence my reluctance in changing its workflow.

On another note, thanks for picking this PR up! I'll do another review pass now, and if it looks good, we can merge it.

docs-partials/builder/common/BlockDevice.mdx
```diff
@@ -16,6 +16,7 @@ launch_block_device_mappings {
     encrypted = true
     kms_key_id = "1a2b3c4d-5e6f-1a2b-3c4d-5e6f1a2b3c4d"
 }
+
```
lbajolet-hashicorp, line 19: I wonder why this changed, is it something that happens when you run `make generate`? If so we're good, I'm more curious than anything

builder/chroot/builder.go
```diff
@@ -376,6 +380,10 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 		errs = packersdk.MultiErrorAppend(errs, errors.New(`The only valid ami_architecture values are "arm64", "i386", "x86_64", or "x86_64_mac"`))
 	}
 
+	if b.config.TpmSupport != "" && b.config.TpmSupport != ec2.TpmSupportValuesV20 {
+		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(`The only valid tpm_support values are %q or the empty string`, ec2.TpmSupportValuesV20))
```
lbajolet-hashicorp, line 384: Little nit that I think I forgot to highlight in my first review, the `or the empty string` part feels off to me, the empty string is the default value, which is equivalent to not specifying it, so I think we can rework this message to give out the only acceptable value.
```suggestion
		errs = packersdk.MultiErrorAppend(errs, fmt.Errorf(`The only valid tpm_support value is %q`, ec2.TpmSupportValuesV20))
```

lbajolet-hashicorp: This would apply to all the builders that support the feature.

lbajolet-hashicorp (approved):
Pre-approving in the expectative for a reroll with my comments addressed, but overall this looks good to me. The only thing I'd maybe like to see would be an acceptance test for `ebssurrogate`, but as it stands I think it's good already, we can do another pass to make sure this works as intended.

Thanks for picking this up again @alemuro !

docs-partials/builder/common/BlockDevice.mdx
alemuro: Yup! I didn't changed the file manually but with `make generate`.

builder/chroot/builder.go
alemuro: Make sense, added to the three builders where this PR apply: `chroot`, `ebssurrogate`, and `instance`. 
Thanks!

alemuro:
Morning @lbajolet-hashicorp , 

I've added tests to the other builders as requested, the `make test` command runs with no failure:

```shell
$ make test
?       github.com/hashicorp/packer-plugin-amazon       [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/chroot        1.088s
ok      github.com/hashicorp/packer-plugin-amazon/builder/common        2.295s
?       github.com/hashicorp/packer-plugin-amazon/builder/common/awserrors      [no test files]
?       github.com/hashicorp/packer-plugin-amazon/builder/common/ssm    [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebs   6.472s
?       github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance        [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebssurrogate  1.100s
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebsvolume     2.581s
ok      github.com/hashicorp/packer-plugin-amazon/builder/instance      2.985s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/ami        0.880s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/parameterstore     2.580s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/secretsmanager     3.102s
?       github.com/hashicorp/packer-plugin-amazon/post-processor/import [no test files]
?       github.com/hashicorp/packer-plugin-amazon/version       [no test files]
```

Regarding the `ebs` builder, I agree 100% with you. 

Thanks for your revision!

alemuro:
Ok I saw the `BlockDevice.mdx` thing. I thought this was generated by the `make generate`, but seems I was completely wrong... I run it again and now the newline has been removed. Apologies 🙏 

Now it should be fine.

docs-partials/builder/common/BlockDevice.mdx
alemuro: Fixed.

lbajolet-hashicorp:
Hi @alemuro,

Thanks for being so quick at integrating my comments, much appreciated!

With this reroll, it all looks good to me, I'm merging this now.

alemuro:
Hello @lbajolet-hashicorp , do you have a tentative date on when this PR will be released on a new release? Thanks!!

lbajolet-hashicorp:
Hi @alemuro,

Considering it's been a few weeks since the last release, and from what I see we don't have a ton on the table on this plugin now, we can probably release a new version soon, in the coming days or maybe next week sounds feasible I would think!

alemuro:
Cool thanks!

PR #380, opened 2023-05-11
.github: add action to run acceptance tests

lbajolet-hashicorp:
Since right now we don't have any way to run acceptance tests on the AWS plugin, we add a workflow that can manually be triggered.

### DIFF
--- a/.github/workflows/acceptance-tests.yml
+++ b/.github/workflows/acceptance-tests.yml
@@ -1,0 +1,75 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+name: "AWS Acceptance Tests"
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "A branch or SHA"
+        default: 'main'
+        required: false
+      run_pattern:
+        description: "The pattern to run tests on"
+        default: "TestAcc"
+        required: false
+
+permissions:
+  contents: read
+
+jobs:
+  get-go-version:
+    runs-on: ubuntu-latest
+    outputs:
+      go-version: ${{ steps.get-go-version.outputs.go-version }}
+    steps:
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3 # v3.5.0
+      - name: 'Determine Go version'
+        id: get-go-version
+        run: |
+          echo "Found Go $(cat .go-version)"
+          echo "go-version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+
+  ebs-acc-tests:
+    needs:
+      - get-go-version
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    name: AWS EBS Acceptance tests
+    steps:
+      - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3 # v3.5.0
+        with:
+          ref: ${{ github.event.inputs.branch }}
+      - uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
+        with:
+          go-version: ${{ needs.get-go-version.outputs.go-version }}
+      - name: "Compile/install the plugin on the current branch"
+        run: |
+          make dev
+      - name: "Install AWS cli and SSM plugin"
+        run: |
+          cd /tmp
+          curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip"
+          unzip awscliv2.zip
+          ./aws/install
+          curl "https://s3.amazonaws.com/session-manager-downloads/plugin/latest/ubuntu_64bit/session-manager-plugin.deb" -o "session-manager-plugin.deb"
+          dpkg -i session-manager-plugin.deb
+          rm -f session-manager-plugin.deb awscliv2.zip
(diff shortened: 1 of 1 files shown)


.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          role-to-assume: "${{ env.AWS_ACC_TEST_ROLE }}"
+          role-session-name: "packer-aws-acceptance-tests"
+          role-duration-seconds: 14400 # 4h
+      - run: |
+          echo "Testing with Go ${{ needs.get-go-version.outputs.go-version }}"
+          PACKER_ACC=1 go test \
+            -timeout 3h \
+            -count 1 \
+            -run "${{ github.event.inputs.run_pattern }}" \
+            ./builder/ebs
```
nywilken, line 66: So a few suggestions here. I would expect the acceptance test to run all test, with the default run pattern to be TestAcc `"*"` doesn't seem to work. 
```suggestion
          PACKER_ACC=1 go test \
            -timeout 3h \
            -count 1 \
            -run "${{ github.event.inputs.run_pattern }}" \
            ./... \
            -v
```

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: "A branch or SHA"
+        default: 'main'
+        required: false
+      run_pattern:
+        description: "The pattern to run tests on"
+        default: "*"
```
nywilken, line 15: Using `"*"` is an invalid regexp expression for go test
```suggestion
        default: "TestAcc"
```

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+name: "AWS Acceptance Tests"
+
+on:
```
nywilken, line 6: Do you plan to add a scheduled cron after running your initial tests?

nywilken (changes requested):

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          echo "go-version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
+
+  ebs-acc-tests:
+    needs:
+      - get-go-version
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    name: AWS EBS Acceptance tests
```
nywilken, line 41: Are you limiting to EBS just for the initial testing?

.github/workflows/acceptance-tests.yml
```diff
@@ -0,0 +1,66 @@
...
+          ref: ${{ github.event.inputs.branch }}
+      - uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
+        with:
+          go-version: ${{ needs.get-go-version.outputs.go-version }}
+      - name: "Compile/install the plugin on the current branch"
+        run: |
+          make dev
+      - uses: aws-actions/configure-aws-credentials@e1e17a757e536f70e52b5a12b2e8d1d1c60e04ef # v2
+        with:
+          aws-access-key-id: "${{ env.AWS_ACC_TEST_KEY_ID }}"
```
nywilken, line 54: Just out of my curiosity, the acceptance service account is used to authenticate and it the assumes another role for testing. Is that correct?

.github/workflows/acceptance-tests.yml
lbajolet-hashicorp: Oh good call for the `aws` cli and plugin, I completely forgot to install them. I'll add this to the workflow

.github/workflows/acceptance-tests.yml
lbajolet-hashicorp: Good call, the pattern was borked indeed, I initially shot for `.*`, but `TestAcc` does the trick indeed.

.github/workflows/acceptance-tests.yml
lbajolet-hashicorp: While I tend to agree, the rationale to separate the actions by builder was so packages could be tested in parallel (although I presume we could do this with an option to `go test` maybe), and so they can have different timeout/token validity periods.
I can change it however, can't hurt to simplify for now and revisit later if the need to split between components arises.

nywilken: That's valid. If you take that approach take care to ensure that there are no tests that overlap with each other or that are sharing the same instance names that would cause any potential conflicts by running in parallel.

lbajolet-hashicorp: True! If we see conflicts arising here we can always rename the instances in the problematic tests too, but we can take a look before to make sure it does work as intended before enabling other builders.

.github/workflows/acceptance-tests.yml
nywilken: Whatever you think is best. I like nightly for active plugins.

lbajolet-hashicorp: I don't mind nightly, but I fear this may become costly to run those every day, while once a week would strike a reasonable balance between costs and safety. We can augment/reduce the frequency if we think it runs too often/not often enough.

nywilken (approved):
LGTM

PR #381, opened 2023-05-12
workflows: run awscli install with update flag

lbajolet-hashicorp:
When running the action to run the acceptance tests, the step to install the AWS cli fails since there's already a version installed.

To prevent this from making the action fail, we invoke the installation script with the `--update' flag, so it will override the current version with the latest from Amazon.


nywilken (approved):
Good with me. Alternative you can remove. I do wonder if there will be a pathing issue.

PR #382, opened 2023-05-12
workflows: remove AWS cli install from acc tests

lbajolet-hashicorp:
Since the Github runner already has the AWS cli and the session-manager plugin installed, there's no point in installing it ourselves, so we remove the script that does that for this workflow.


nywilken (approved):
👍 

https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#cli-tools

PR #383, opened 2023-05-12
workflows: reference secrets not as env vars

lbajolet-hashicorp:
The action to get temporary credentials for AWS was referring to secrets under `env', which resulted in the arguments for the action being empty, and therefore failing.


nywilken (approved):
LGTM

PR #388, opened 2023-05-16
feat: add ignore failure in ami datasource

### LINKED ISSUE #387: Ignore failure flag when ami datasource can't find ami
#### Description
I would like to add an option for ami datasource that will "ignore failure" when a ami is not found between aws ami's.

#### Use Case(s)

I have 2 cases using same packer file:
- the ami exists
- the ami does not exits, and I want to fall back to hardcoded ami id.
The current version of the plugin fails when the datasource doesn't find any ami by the given filters. 

#### Potential References
Error example:
`Error: Datasource.Execute failed: No AMI was found matching filters:`

I will apply a PR for this one and mention it here.

galbirk:
1. Added `ignore_failure` flag to ami datasource:
- if `true` and the image not found -> returns `nil`
- else - same as current behavior.

Defaulted to `false`.

2. Some formatting required to pass github actions test.

Tests results:
``` 
$ make test
?       github.com/hashicorp/packer-plugin-amazon       [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/chroot        3.469s
ok      github.com/hashicorp/packer-plugin-amazon/builder/common        2.035s
?       github.com/hashicorp/packer-plugin-amazon/builder/common/awserrors      [no test files]
?       github.com/hashicorp/packer-plugin-amazon/builder/common/ssm    [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebs   13.905s
?       github.com/hashicorp/packer-plugin-amazon/builder/ebs/acceptance        [no test files]
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebssurrogate  4.812s
ok      github.com/hashicorp/packer-plugin-amazon/builder/ebsvolume     3.097s
ok      github.com/hashicorp/packer-plugin-amazon/builder/instance      4.228s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/ami        2.356s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/parameterstore     3.011s
ok      github.com/hashicorp/packer-plugin-amazon/datasource/secretsmanager     2.710s
?       github.com/hashicorp/packer-plugin-amazon/post-processor/import [no test files]
?       github.com/hashicorp/packer-plugin-amazon/version       [no test files]
```

Closes #387

### DIFF
--- a/builder/chroot/step_attach_volume.go
+++ b/builder/chroot/step_attach_volume.go
@@ -19,8 +19,9 @@ import (
 // available device location.
 //
 // Produces:
-//   device string - The location where the volume was attached.
-//   attach_cleanup CleanupFunc
+//
+//	device string - The location where the volume was attached.
+//	attach_cleanup CleanupFunc
 type StepAttachVolume struct {
 	PollingConfig *awscommon.AWSPollingConfig
 	attached      bool
--- a/builder/chroot/step_create_volume.go
+++ b/builder/chroot/step_create_volume.go
@@ -22,7 +22,8 @@ import (
 // device of the AMI.
 //
 // Produces:
-//   volume_id string - The ID of the created volume
+//
+//	volume_id string - The ID of the created volume
 type StepCreateVolume struct {
 	PollingConfig         *awscommon.AWSPollingConfig
 	volumeId              string
--- a/builder/chroot/step_flock.go
+++ b/builder/chroot/step_flock.go
@@ -17,7 +17,8 @@ import (
 // StepFlock provisions the instance within a chroot.
 //
 // Produces:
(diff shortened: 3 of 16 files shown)


nywilken:
Hi @galbirk thanks for opening up this PR and its corresponding issue. This particular solution while valid feels like a one off solution to a larger workflow problem around providing fallbacks when a data sources returns a failure. I'm going to close this  pull-request as invalid, as this is not the approach we want to take. But I will keep the issue open because there is a workflow discussion we can have for supporting conditional to handle within a Packer template. 

I will bring this issue internal to see if there is a workaround or potential workflow we can leverage at this time.

PR #390, opened 2023-05-24
Acc tests schedule

lbajolet-hashicorp:
This PR adds a schedule to run acceptance tests every Sunday at 4:00 AM (UTC), and adds a job to run acceptance tests on the `ebssurrogate` builder as well.


nywilken (approved):
LGTM

PR #395, opened 2023-05-31
fixing text for IAM permission needed

elreydetoda:
There was an error for in the Visual Policy Editor when I copy/pasted the JSON and then went to view it, picture below.

image

So, I found that it was a capitalization error with that permission, so I changed `ec2:CreateKeypair` to `ec2:CreateKeyPair`. (capital `p` in the `Pair` part of the word)


lbajolet-hashicorp (approved):
Hey @elreydetoda,

Good catch, thank you for this fix!

I'm merging this now.

Issue #398, opened 2023-06-09
temporary_iam_instance_profile_policy_document doesn't use poller configs


imcdo:
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue
I have configured a `temporary_iam_instance_profile_policy_document` for an `amazon-ebs` source.  However it always fails to create the role with 
```
Timed out waiting for temporary role
```
it seems like instead of passing forward the global polling configs such that this is configurable, we are doing:
https://github.com/hashicorp/packer-plugin-amazon/blob/8e1e361e8d32e909f6fc8597625f7a0d95c088d0/builder/common/step_iam_instance_profile.go#LL119C16-L119C35

which calls `WaitUntilRoleExistsWithContext` with its default values of 20 attempts with a 1 second backoff instead we should pass in the poller config and allow users to configure this

https://github.com/aws/aws-sdk-go/blob/a5bef1542952d979233bd92569aa09eeb7686c6a/service/iam/waiters.go#L133-L134

#### Reproduction Steps
set `temporary_iam_instance_profile_policy_document`
and have it take more than 30 seconds for aws to create the role.

Steps to reproduce this issue

### Plugin and Packer version
From `packer version`
`1.0.9`

### Simplified Packer Buildfile

temporary_iam_instance_profile_policy_document {
    Statement {
      Action = [
        "ecr:GetAuthorizationToken",
        "ecr:DescribeImages",
        "ecr:ListImages",
        "ecr:BatchGetImage",
        "ecr:DescribeRegistry",
        "ecr:DescribeRepositories",
        "ecr:ListTagsForResource",
        "ecr:GetDownloadUrlForLayer"
      ]
      Effect   = "Allow"
      Resource = ["*"]
    }
    Version = "2012-10-17"
  }
### Operating system and Environment details

OS, Architecture, and any other information you can provide about the
environment.
ubuntu20
### Log Fragments and crash.log files

Include appropriate log fragments. If the log is longer than a few dozen lines,
please include the URL to the [gist](https://gist.github.com/) of the log or
use the [Github detailed format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) instead of posting it directly in the issue.

```
name.amazon-ebs.ubuntu: output will be in this color. 02:10
==> name.amazon-ebs.ubuntu: Prevalidating any provided VPC information 02:10
[...]

PR #399, opened 2023-06-09
bug: Pass Polling Config to iam instance profile step

imcdo:
Pass in the global polling config into the iam instance profile step.  Do this so configurations get used when trying to create a temporary iam role for the builder.
Closes [#398](https://github.com/hashicorp/packer-plugin-amazon/issues/398)


lbajolet-hashicorp (approved):
LGTM, thanks for the fix @imcdo!

PR #404, opened 2023-07-13
common: add include_deprecated option to filters

### LINKED ISSUE #403: Support deprecated sources by adding a `IncludeDeprecated` flag to `source_ami_filter`
#### Community Note

#### Description

A call to AWS API `DescribeImages` can contain a boolean parameter `IncludeDeprecated`. Sending a `DescribeImages` with this flag to `True` while using Filters to find an image will add in the response non-private images which have been marked as deprecated. 

I think this parameter is useful and should be supported by packer-plugin-amazon. 

#### Use Case(s)

Using deprecated images as source.
Issue I'm facing right now: AMI was deprecated and we can't build from this images anymore. I'd like to only acknowledge and use `IncludeDeprecated` instead of the alternative (manually calling `DescribeImage` before the call to packer and find the ami-id to pass to packer).

#### Potential References
DescribeImage AWS API doc: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeImages.html
DescribeImage golang sdk: https://github.com/aws/aws-sdk-go-v2/blob/38e10609d10442e497844d3add2e4b9c6d2889f1/service/ec2/api_op_DescribeImages.go#L128C3-L128C20

Implementation:
AmiFilterOptions https://github.com/hashicorp/packer-plugin-amazon/blob/877a7b15d179e840e1e3f191644c0eeb362551cb/builder/common/ami_filter.go#L14
DescribeImage parameters set: https://github.com/hashicorp/packer-plugin-amazon/blob/877a7b15d179e840e1e3f191644c0eeb362551cb/builder/common/ami_filter.go#L49-L59
DescribeImage request: https://github.com/hashicorp/packer-plugin-amazon/blob/877a7b15d179e840e1e3f191644c0eeb362551cb/builder/common/step_pre_validate.go#L103-L108

lbajolet-hashicorp:
When getting the AMIs to build a new image on, we never included deprecated AMIs, as is the default when describing images on AWS.

However, the request does offer this option, so we add this to the options to pick from when getting the base AMI.

Closes #403

### DIFF
--- a/.web-docs/components/data-source/ami/README.md
+++ b/.web-docs/components/data-source/ami/README.md
@@ -38,6 +38,10 @@ This selects the most recent Ubuntu 16.04 HVM EBS AMI from Canonical. Note that 
 - `most_recent` (bool) - Selects the newest created image when true.
   This is most useful for selecting a daily distro build.
 
+- `include_deprecated` (bool) - Include deprecated AMIs in the filtered response. Defaults to false.
+  If you are the AMI owner, deprecated AMIs appear in the response
+  regardless of what is specified for `include_deprecated`.
+
 <!-- End of code generated from the comments of the AmiFilterOptions struct in builder/common/ami_filter.go; -->
 
 
--- a/builder/common/ami_filter.go
+++ b/builder/common/ami_filter.go
@@ -25,6 +25,10 @@ type AmiFilterOptions struct {
 	// Selects the newest created image when true.
 	// This is most useful for selecting a daily distro build.
 	MostRecent bool `mapstructure:"most_recent"`
+	// Include deprecated AMIs in the filtered response. Defaults to false.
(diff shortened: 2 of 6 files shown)


Ckarles:
Thanks for the quick update!

Should there be an implementation on top of this to add explanatory warnings / errors which would help the maintenance and troubleshooting ?

For example:
- By default, filter with include_deprecated.
- If the image chosen is deprecated (`.DeprecationTime < now()`):
  - if not `include_deprecated`: return an error: image was found for this filter, but deprecated, use `include_deprecated` if you'd like to use it
  - if `include_deprecated`: display a warning: "using a deprecated image", ignoring because of `include_deprecated=True`

wdyt ?

nywilken (approved):
I left a question, possible suggestion, but this is otherwise good to me.

PR #412, opened 2023-08-11
ebs: skip_create_ami skips prevalidation of AMI name (#409)

### LINKED ISSUE #409: Enabling skip_create_ami should skip prevalidation of the AMI Name in amazon-ebs builder
#### Description

Enabling **skip_create_ami** in the in amazon-ebs builder unnecessarily prevalidates the AMI name. This is an annoyance for development since I have to delete an existing AMI for a build that will not create/replace the AMI. 

#### Expected Behavior
Enabling **skip_create_ami** should skip the AMI name prevalidation similar to the **force_deregister** and **skip_save_build_region** options.

#### Plugin and Packer version
Packer v1.9.2
amazon plugin : 1.2.6

#### Use Case(s)

Development builds in Packer

#### Workaround
Enable **force_deregister** which forces the prevalidation to be skipped, however this comes at the expense of deregistering the existing AMI at the end :frowning_face:

lillarspillars:
Pretty simple fix that just adds a conditional to the **step_pre_validate** to skip it in case the skip_create_ami flag is set to true. I kept the look and format consistent with the existing skips that were already present.

Closes #409

### DIFF
--- a/builder/common/step_pre_validate.go
+++ b/builder/common/step_pre_validate.go
@@ -24,6 +24,7 @@ type StepPreValidate struct {
 	DestAmiName        string
 	ForceDeregister    bool
 	AMISkipBuildRegion bool
+	AMISkipCreateImage bool
 	VpcId              string
 	SubnetId           string
 	HasSubnetFilter    bool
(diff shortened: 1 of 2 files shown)


nywilken (approved):
LGTM thanks for the fix.

PR #414, opened 2023-09-11
ebs: don't error on AMI sharing test unset

lbajolet-hashicorp:
Testing for the AMI sharing between accounts requires multiple extra variables to be set so we can interpolate those values in the template we run the test on.

In case it is not set, the current setup makes acceptance tests fail, which can be counter-productive.
To avoid this problem, we don't error when these variables are not set, but instead we skip the test.


nywilken (approved):
LGTM

PR #430, opened 2023-11-03
Replace Xenial with Jammy image

nywilken:
This change updates the Ubuntu image in the Linux acceptance test to a non-deprecated Jammy 22.04 server. This change fixes the failing acceptance tests.


lbajolet-hashicorp (approved):
Yes and yes! Thanks for that @nywilken

Issue #445, opened 2023-12-19
Enabling fast_launch on Windows AMI does not work


tivanov-qb:
When filing a bug, please include the following headings if possible. Any
example text in this template can be deleted.

#### Overview of the Issue

After adding fast_launch to my packer builders block, AMIs I'm generating do not have the fast launch enabled.
  "builders": [
    {
      "fast_launch": {
        "enable_fast_launch": true,
        "target_resource_count": 1,
        "max_parallel_launches": 6
      }
}

I'm using latest packer and plugin versions.
The AMI is created successfully, however the fast launch is not enabled.

tivanov-qb:
@lbajolet-hashicorp

lbajolet-hashicorp:
Hi @tivanov-qb,

Thanks for reporting this, could I ask you to share logs of what happens when creating the image? Normally with fast launch enabled, there should be a step that creates the launch template, and enables fast launch for them, and this step has a few logs that can help understanding what goes wrong here.

We wait until the state of the image changes to either `enabled`, `enabled-failed`, or `enabling-failed`, but I do see that we don't check for the state of the fast-launch after it's done waiting on the snapshots' creation.
In your case I suspect this could be due to a problem here. Have you sysprep'd the image by any chance? This is required to enable fast-launch.

I'll see to add this warning/error when we try to enable fast-launch. Maybe we should fail the build if that doesn't work.

lbajolet-hashicorp:
Also, IIRC you can get information about what failed regarding the `fast-launch` enablement on the AWS console for the attempt, I'd suggest taking a look over there (sorry I haven't worked on this in a while so my memory's a bit nebulous)

tivanov-qb:
hey @lbajolet-hashicorp ! thank you so much for the swift response! The logs are not indicating anything regarding the fast launch and I'm running the packer build with debug. I do not see any errors at the time of building the AMI. Unfortunately, after the AMI is ready, the fast launch is shown as disabled and I can enable it manually from the AWS console :(
So no enabled, enabled-failed, or enabling-failed statuses or indication for such.
I do use sysprep for each AMI that I create. I wonder what I'm doing wrong and why this change is not reflected into my AMI creation :(

lbajolet-hashicorp:
@tivanov-qb if you don't sysprep your instances then this is likely why `fast-launch` fails to enable, it's likely you end-up in one of the other states and why the snapshots aren't created for launching your instances later.

[...]

PR #446, opened 2023-12-20
Add more diags on fast-launch enable step

lbajolet-hashicorp:
Following up on issue #445, this PR adds more diagnostics to the step that enables fast-launch.

Since the wait step only waited for the state to change to either "enabled", "enabled-failed" or "enabling-failed", but later on the code didn't check the state was indeed "enabled", we could end-up in a situation where the fast-launch step failed on AWS, but the plugin would count it as a success.
Note: I couldn't replicate such a scenario in my tests, but this remains a possibility, so I hope the community will someday share a template that we can adapt to ensure the plugin does take this into account.

Also, added some checks to signal the step is being skipped if no AMI's present in the plugin's state bag, as otherwise the step would silently succeed without doing anything.

Finally, we also fix the acceptance tests, as the AMI we picked for them did not exist anymore, and therefore would always fail.

### DIFF
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -1760,9 +1760,17 @@ build {
 `
 
 const testWindowsFastBoot = `
+data "amazon-ami" "windows-ami" {
+	filters = {
+		name = "Windows_Server-2016-English-Core-Base-*"
+	}
+	owners = ["801119661308"]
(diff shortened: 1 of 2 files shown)


JenGoldstrich (approved):
LGTM! I think we should add a fast launch unit test eventually as I mentioned offline but I don't think its blocking this

PR #451, opened 2024-01-22
ebs: add fast-launch support with AMI copies

lbajolet-hashicorp:
Prior to this commit, enabling fast-launch was done before copying the AMI to other regions, if the option was selected.

This was problematic, as the pre-provisioned snapshots produced by enabling fast-launch weren't copied over to other regions, so only the source image was supporting fast-launch.

To remedy this problem, we now enable fast-launch after the image is copies to other regions. This fixes the problem, but takes more time to complete the build, since each enabling takes time.

Also, this requirement means that each region may have its own launch template to use for enabling fast-launch, so we introduce a new block in the configuration to specify which region should use which template.
This option does not directly supersede the original `fast_launch_template*` options, but instead assumes that if specified,
will only apply to the region the AMI is built in, and other regions will default if unspecified.

Closes: #445

### DIFF
--- a/.web-docs/components/builder/ebs/README.md
+++ b/.web-docs/components/builder/ebs/README.md
@@ -1671,24 +1671,43 @@ https://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/win-ami-config-fast-launc
 
 - `enable_fast_launch` (bool) - Configure fast-launch for Windows AMIs
 
-- `template_id` (string) - The ID of the launch template to use for the fast launch
+- `template_id` (string) - The ID of the launch template to use for fast launch for the main AMI.
   
   This cannot be specified in conjunction with the template name.
   
   If no template is specified, the default launch template will be used,
   as specified in the AWS docs.
+  
+  If you copy the AMI to other regions, this option should not
+  be used, use instead the `fast_launch_template_config` option.
 
-- `template_name` (string) - The name of the launch template to use for fast launch
+- `template_name` (string) - The name of the launch template to use for fast launch for the main AMI.
   
   This cannot be specified in conjunction with the template ID.
   
   If no template is specified, the default launch template will be used,
   as specified in the AWS docs.
+  
+  If you copy the AMI to other regions, this option should not
+  be used, use instead the `fast_launch_template_config` option.
 
-- `template_version` (int) - The version of the launch template to use
+- `template_version` (int) - The version of the launch template to use for fast launch for the main AMI.
   
   If unspecified, and a template is referenced, this will default to
   the latest version available for the template.
+  
+  If you copy the AMI to other regions, this option should not
+  be used, use instead the `fast_launch_template_config` option.
+
+- `region_launch_templates` ([]FastLaunchTemplateConfig) - RegionLaunchTemplates is the list of launch templates per region.
+  
+  This should be specified if you want to use a custom launch
+  template for your fast-launched images, and you are copying
+  the image to other regions.
+  
+  Note: all the regions don't need an entry in this map, but if you
+  don't specify a region's template, a default one will be picked
+  by AWS.
 
 - `max_parallel_launches` (int) - Maximum number of instances to launch for creating pre-provisioned snapshots
   
--- a/builder/ebs/builder.go
+++ b/builder/ebs/builder.go
@@ -145,7 +145,24 @@ func (b *Builder) Prepare(raws ...interface{}) ([]string, []string, error) {
 	errs = packersdk.MultiErrorAppend(errs, b.config.LaunchMappings.Prepare(&b.config.ctx)...)
 	errs = packersdk.MultiErrorAppend(errs, b.config.RunConfig.Prepare(&b.config.ctx)...)
 
+	b.config.FastLaunch.defaultRegion = b.config.RawRegion
 	errs = packersdk.MultiErrorAppend(errs, b.config.FastLaunch.Prepare()...)
+	for _, templateConfig := range b.config.FastLaunch.RegionLaunchTemplates {
+		exists := false
+		for _, cpRegion := range b.config.AMIRegions {
+			if cpRegion == templateConfig.Region {
+				exists = true
+				break
+			}
+		}
+		if b.config.RawRegion == templateConfig.Region {
+			exists = true
+		}
+
+		if !exists {
+			errs = packersdk.MultiErrorAppend(errs, fmt.Errorf("Launch template specified for enabling fast-launch on region %q, but the AMI won't be copied there.", templateConfig.Region))
+		}
+	}
 
 	if b.config.IsSpotInstance() && (b.config.AMIENASupport.True() || b.config.AMISriovNetSupport) {
 		errs = packersdk.MultiErrorAppend(errs,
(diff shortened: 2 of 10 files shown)


lbajolet-hashicorp:
Note: I'll try to add some unit-tests for the configurations before we can merge.

@tivanov-qb: if possible, could I ask you to test this fix on your configuration? I'd like to have a confirmation that this does fix your problem, even if some acceptance tests point in the right direction, we can't be too careful :)

JenGoldstrich (approved):
This LGTM minus a minor test nit

tivanov-qb:
@lbajolet-hashicorp looking forward for this to be merged, so I can kick off a test. Thank you!

lbajolet-hashicorp:
@tivanov-qb any chance you can build the plugin and test it locally on one of your configurations? I'd prefer if you could test it before we release it since we'd have to crank out another release quickly after if that change doesn't fix your issue

tivanov-qb:
> @tivanov-qb any chance you can build the plugin and test it locally on one of your configurations? I'd prefer if you could test it before we release it since we'd have to crank out another release quickly after if that change doesn't fix your issue

@lbajolet-hashicorp  absolutely, could you please share how to point to branch/tag ?
this is how i'm using the plugin (hcl format)

packer {
  required_plugins {
    amazon = {
      version = "~> 1"
    }
  }
}

tivanov-qb:
> At the moment unfortunately you won't be able to use a `required_plugins` with a manually installed plugin, we're working on supporting this workflow right now, but it's not been released yet.
> 
> I would advise you to remove the `required_plugins` block from your template, and store the built version of the plugin for your OS/arch alongside the working directory you invoke Packer from. Also I would suggest invoking packer with `PACKER_LOG=1` to make sure the right plugin binary is invoked for the test.
> 
> If you need more guidance I suggest referring to the docs on [installing plugins](https://developer.hashicorp.com/packer/docs/plugins/install-plugins#installation-guides), especially under the `Manual installation` tab. Let me know if you need some help with that.

sorry.. but I think I will need help with that.. the document says replace the binary, but how do I get the testing binary to replace it with the actual one ?

lbajolet-hashicorp:
In this case you'll have to build it; if you clone the repository and checkout the branch in which I made those changes, you'll be able to build the plugin and move it where needed.

You'll need the Go toolchain for that, and possibly `make` if you want to run the commands through the provided `Makefile`, though for something as simple as this build, you can directly invoke `go build -o packer-plugin-amazon`, which will build the plugin for your environment.
At that point, you should be able to use it with Packer

tivanov-qb:
@lbajolet-hashicorp i already wasted couple of hours and this just does not work. if the plugin is called packer-plugin-amazon is not being picked by packer.. if the name is packer-plugin-amazon_v1_linux_amd64 it fails because it cannot find a checksum file..

tivanov-qb:
```
[10:23:04  2024/01/24 08:23:04 Detected config directory from env var: /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config
10:23:04  2024/01/24 08:23:04 [TRACE] discovering plugins in /bin
10:23:04  2024/01/24 08:23:04 [TRACE] discovering plugins in .
10:23:04  2024/01/24 08:23:04 [TRACE] discovering plugins in /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins
10:23:04  2024/01/24 08:23:04 [INFO] Discovered potential plugin: amazon = /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon_linux_amd64
10:23:04   Error loading configuration: 
10:23:04  
10:23:04  fork/exec /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon_linux_amd64: no such file or directory](url)
```

tivanov-qb:
I give up, i don't know how to workaround/fix this error 

`11:11:13  2024/01/24 09:11:13 [INFO] Packer version: 1.10.0 [go1.20.11 linux amd64]
11:11:13  2024/01/24 09:11:13 [TRACE] discovering plugins in /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins
11:11:13  2024/01/24 09:11:13 [INFO] Discovered potential plugin: amazon = /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon
11:11:13   Error loading configuration: 
11:11:13  
11:11:13  fork/exec /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon: no such file or directory`

lbajolet-hashicorp:
> I give up, i don't know how to workaround/fix this error
> 
> ```
> `11:11:13  2024/01/24 09:11:13 [INFO] Packer version: 1.10.0 [go1.20.11 linux amd64]
> 11:11:13  2024/01/24 09:11:13 [TRACE] discovering plugins in /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins
> 11:11:13  2024/01/24 09:11:13 [INFO] Discovered potential plugin: amazon = /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon
> 11:11:13   Error loading configuration: 
> 11:11:13  
> 11:11:13  fork/exec /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon: no such file or directory`
> ```

This does look weird. Is this the binary you built? I'm surprised it cannot start it at all, can you run it from your shell at least? If you try to run ` /mnt/ssd/workspace/workspace/1_EnableFastLaunchForWindowsAMIs/packer_config/.packer.d/plugins/packer-plugin-amazon describe`, does it return the description of the plugin?

Let's try to figure out the source of the problem if possible, there could be some other bug lurking around that we can maybe fix by the next release.

In the meantime, I'll address @JenGoldstrich's comment and merge this PR so we can release the plugin, hopefully the problem's fixed and you can roll with it then.

tivanov-qb:
@lbajolet-hashicorp seems you fixed it! AMIs are now created with fast launch enabled and same goes for the copied AMIs. Great work! Thank you!

tivanov-qb:
@lbajolet-hashicorp i just noticed that AMIs shared via org arn are not actually having the fast launch enabled. so in my source ci account all looks good, but fast launch is not enabled in the destination accounts :(

tivanov-qb:
might not be a surprise for you, but aws confirmed fast launch setting is not automatically carried over to the destination accounts

lbajolet-hashicorp:
Hi @tivanov-qb,

This doesn't surprise me that much since fast-launch basically pre-provisions snapshots for deploying the AMI with, which AFAIK are account-bound, so that checks out that for other accounts, `fast-launch` won't be exported.

I don't think this is something we can support in Packer tbh.

PR #453, opened 2024-01-29
docs: fix fast-launch-template-config docs

lbajolet-hashicorp:
Introduced with v1.3.0 of the plugin, the fast-launch-template-config was not added to the docs, and since the partials were generated, but not committed, the release docs deployment failed.

This likely hints at the checks for the PR not being sufficient, as they should have pointed it out, but that wasn't the case.

Follow-up to: #451


nywilken (approved):
LGTM

PR #479, opened 2024-04-24
common: use regional STS on non-default regions

### LINKED ISSUE #469: Unable to copy images to non-default regions with AssumeRole
#### Overview of the Issue

I'm able to successfully assume a role and copy images to multiple regions that are enabled by default in AWS. If I try to copy a region to a non-default region, for example Hong Kong, I will get the following error 

```
* Error Copying AMI (ami-0d68130680fad52b4) to region (ap-east-1): AuthFailure: AWS was not able to validate the provided access credentials
```

It seems that Packer is calling the global STS endpoint which issues version 1 tokens which can not be used in non-default regions, see https://repost.aws/knowledge-center/iam-validate-access-credentials

#### Reproduction Steps

Create a configuration with AssumeRole and attempt to copy an image to a non-default region, ie one you must enable explicitly like Hong Kong (ap-east-1)

### Plugin and Packer version

Latest and 1.3.1

Packer should use the regional STS endpoint to ensure version 2 tokens are used

lbajolet-hashicorp:
STS is the service used by AWS for emitting authentication tokens for API clients.

This comes in two variants: v1 (global) and v2 (regional). As of today (2024-04-24), the default for the Go SDK is "legacy", i.e. if the connection is used to communicate with a non-default region it will use a regional endpoint, otherwise it'll use the global endpoint.

Builds are generally not affected by operations like these as the SDK will pick the right type of endpoint for that, but problems may arise later, when copying AMIs for example, as they will need tokens compatible with both the source and destination regions.

This means that if the build was performed in a default region, then copied to a non-default region, we'll have gotten a v1 (global) token, which will be rejected by the target region, causing the build to fail.

This is already fixable by user-action, through either a setting in their AWS config file, or through an environment variable, but this may come as a surprise if users aren't aware of that pitfall.

Therefore, this commit attempts to heuristically determine if an action may fail in the process, and enable regional endpoints for the EC2 session we create during a build.

Note: the volume builder and the post-processor are not affected by this, as they only work within one region at a time, so the SDK will choose the right type of endpoint/token for the action, and no cross-region action will be done.

Closes #469


williamb1024:
I believe you meant @william00179

lbajolet-hashicorp:
Oh. My bad @williamb1024, wrong William indeed, sorry about the unwarranted ping.

PR #481, opened 2024-05-10
feat: allow customizing spot allocation strategy

### LINKED ISSUE #427: Allow Customizing Spot Allocation Strategy
#### Community Note

#### Description

EC2 allows choosing multiple [allocation strategies](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html) for spot requests. It would be great to be able to specify the Allocation strategy for packer builds on Spot.

#### Use Case(s)

In our use case (building AMIs in CI/CD) we are seeing frequent evictions of spot instances mid-build, or even before launching the build instances. It could help to choose a strategy like `capacity-optimized` or `price-capacity-optimized`, in order to see less frequent evictions and failures due to lack of capacity.

#### Potential configuration

An option that exposes `spot_allocation_strategy` on the AWS builder would do the trick. That option could still default to `lowest-price` for backwards behavioral compatibility. But `price-capacity-optimized` is the recommended value.

#### Potential References

It looks like the AWS SDK exposes this configuration point as an Enum (sorry, this file is to big to permalink to):
https://github.com/aws/aws-sdk-go/blob/main/service/ec2/api.go

I think it would only be a matter of adding that option to the spot fleet request here?
https://github.com/hashicorp/packer-plugin-amazon/blob/e8a4526665223b6cf2830f47f0fb31baab180075/builder/common/step_run_spot_instance.go#L368-L384

From there, a configuration option would need to be exposed from input and threaded to the fleet request.

---

Apologies if this is too vague, as i'm not very well versed in Go.

vucong2409:
Allow user to pass Spot Allocation Strategy into Fleet Request. 

Closes #427

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -843,11 +843,10 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
-	if c.SpotAllocationStrategy != "" {
-		if !slices.Contains(ec2.SpotAllocationStrategy_Values(), c.SpotAllocationStrategy) {
-			errs = append(errs, fmt.Errorf(
-				"Unknown spot_allocation_strategy: %s", c.SpotAllocationStrategy))
-		}
+	if c.SpotAllocationStrategy != "" && !slices.Contains(ec2.SpotAllocationStrategy_Values(),
+		c.SpotAllocationStrategy) {
+		errs = append(errs, fmt.Errorf(
+			"Unknown spot_allocation_strategy: %s", c.SpotAllocationStrategy))
 	}
 
 	if c.UserData != "" && c.UserDataFile != "" {
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -404,10 +404,10 @@ type RunConfig struct {
 	//   filter, but will cause Packer to fail if the `source_ami` does not exist.
 	SourceAmiFilter AmiFilterOptions `mapstructure:"source_ami_filter" required:"false"`
 	// One of  `price-capacity-optimized`, `capacity-optimized`, `diversified` or `lowest-price`.
-	// 	 The strategy that determines how to allocate the target Spot Instance capacity
-	// 	 across the Spot Instance pools specified by the EC2 Fleet launch configuration.
-	// 	 If this option is not set/set to invalid value, Packer will use default option provided by the SDK (currently `lowest-price`).
-	// 	 For more information, see [Amazon EC2 User Guide] (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html)
+	// The strategy that determines how to allocate the target Spot Instance capacity
+	// across the Spot Instance pools specified by the EC2 Fleet launch configuration.
+	// If this option is not set, Packer will use default option provided by the SDK (currently `lowest-price`).
+	// For more information, see [Amazon EC2 User Guide] (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html)
 	SpotAllocationStrategy string `mapstructure:"spot_allocation_strategy" required:"false"`
 	// a list of acceptable instance
 	// types to run your build on. We will request a spot instance using the max
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -11,9 +11,11 @@ import (
 	"net"
 	"os"
 	"regexp"
+	"slices"
 	"strings"
 	"time"
 
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/hashicorp/packer-plugin-sdk/communicator"
 	"github.com/hashicorp/packer-plugin-sdk/template/config"
 	"github.com/hashicorp/packer-plugin-sdk/template/interpolate"
@@ -838,6 +840,13 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		if c.SpotPrice == "" || c.SpotPrice == "0" {
 			errs = append(errs, fmt.Errorf(
 				"spot_tags should not be set when not requesting a spot instance"))
+		}
+	}
+
+	if c.SpotAllocationStrategy != "" {
+		if !slices.Contains(ec2.SpotAllocationStrategy_Values(), c.SpotAllocationStrategy) {
+			errs = append(errs, fmt.Errorf(
+				"Unknown spot_allocation_strategy: %s", c.SpotAllocationStrategy))
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


builder/common/run_config.go
```diff
@@ -402,6 +402,12 @@ type RunConfig struct {
 	//   criteria provided in `source_ami_filter`; this pins the AMI returned by the
 	//   filter, but will cause Packer to fail if the `source_ami` does not exist.
 	SourceAmiFilter AmiFilterOptions `mapstructure:"source_ami_filter" required:"false"`
+	// One of  `price-capacity-optimized`, `capacity-optimized`, `diversified` or `lowest-price`.
+	// 	 The strategy that determines how to allocate the target Spot Instance capacity
+	// 	 across the Spot Instance pools specified by the EC2 Fleet launch configuration.
+	// 	 If this option is not set/set to invalid value, Packer will use default option provided by the SDK (currently `lowest-price`).
+	// 	 For more information, see [Amazon EC2 User Guide] (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-fleet-allocation-strategy.html)
```
lbajolet-hashicorp, line 409: I suggest removing the superfluous tabs in this section in order to be uniform vs. the other sections documented sections

builder/common/step_run_spot_instance.go
```diff
@@ -383,6 +385,12 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
 		Type: aws.String("instant"),
 	}
 
+	if slices.Contains(ec2.SpotAllocationStrategy_Values(), s.SpotAllocationStrategy) {
```
lbajolet-hashicorp, line 388: The idea makes sense, but I would suggest doing the check in the config's Prepare function, and error if the allocation strategy is not empty and not among the supported values.
Once that check is implemented over here, this can become a simple `if s.SpotAllocationStrategy != ""`.

What worries me with the current implementation is that if someone defines an invalid value for any reason (typo typically comes to mind), Packer will not warn/error of this, and will silently ignore the invalid configuration value, which is something we try to avoid as much as possible.

lbajolet-hashicorp (changes requested):
Hi @vucong2409,

Thanks for taking a jab at this!

I left a couple comments/suggestions regarding the formatting of the docs and validation of the option, please let me know if you need some help or more explanation, otherwise feel free to address this feedback and I'll do another round of review then.

Thanks again!

builder/common/run_config.go
vucong2409: Fixed in e2ed97d.

builder/common/step_run_spot_instance.go
vucong2409: Thanks for pointing out! I've moved that check to prepare step in 807b998. However, should I use the `slice.contains` with the `_Values` method from `ec2` library like the current version, or should change it to a manual string/array check like other checks (for code consistency)?

builder/ebs/builder.go
```diff
@@ -251,6 +251,7 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			SpotTags:                          b.config.SpotTags,
 			Tags:                              b.config.RunTags,
 			SpotInstanceTypes:                 b.config.SpotInstanceTypes,
+			SpotAllocationStrategy:            b.config.RunConfig.SpotAllocationStrategy,
```
lbajolet-hashicorp, line 254: I believe this setting should be replicated to other builders, assuming this is compatible with them.
Otherwise I would suggest either:

1. moving that argument to `builder/ebs/config.go` so it is not common
2. checking that it is not set in the configs for the builders that aren't compatible with it, this can be done in the `Prepare` function for each builder.

builder/common/run_config.go
```diff
@@ -843,6 +851,13 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		}
 	}
 
+	if c.SpotAllocationStrategy != "" {
+		if !slices.Contains(ec2.SpotAllocationStrategy_Values(), c.SpotAllocationStrategy) {
```
lbajolet-hashicorp, line 855: I believe both `if`s can be mutualised
```suggestion
		if c.SpotAllocationStrategy != "" && !slices.Contains(ec2.SpotAllocationStrategy_Values(), c.SpotAllocationStrategy) {
```

lbajolet-hashicorp: Also to reply to your question in a comment from a previous review: using `slices.Contains` is fine by my standards! No need to make it manual, I suspect those were made this way in other cases because either the function did not exist at the time it was added, or whomever implemented it wasn't aware it existed, functionally it is essentially the same, might as well use what the library offers us.

builder/ebs/builder.go
lbajolet-hashicorp: Note: if the `ebs` builder's the only one to support spot instances, I would suggest going with approach 1

lbajolet-hashicorp (commented):
Hi @vucong2409,

Sorry I forgot to do another pass of review after you updated this, I've left another review on the code, this looks good however I'm not sure the option should be common if it is ebs-specific, please let me know if something's unclear in my comments, and I can try to clarify.

Otherwise I'll let you address those comments, and I'll be back later for another round of reviews.

Thanks!

njloof:
Wondering if we can resurrect this change. @vucong2409 do you have bandwidth to finish this up?

builder/ebs/builder.go
kp2099: I have added the setting to all the other builders that support Spot Instances.

builder/common/run_config.go
kp2099: Addressed it.

builder/common/run_config_test.go
```diff
@@ -582,3 +582,13 @@ func TestRunConfigPrepare_WithCapacityReservations(t *testing.T) {
...
 	}
 }
+
+func TestRunConfigPrepare_EnableSpotInstanceBadSpotAllocationStrategy(t *testing.T) {
+	c := testConfig()
+	// There should be some error when Spot Allocation Strategy is invalid.
+	c.SpotAllocationStrategy = "very-expensive-one"
+	err := c.Prepare(nil)
+	if len(err) != 1 {
+		t.Fatalf("Should error if spot_allocation_strategy is invalid.")
```
mogrogan, line 592: nit: not really important but in this case I would output the number of error received like (ie: expected a single error, got 0|2|3... )

mogrogan (approved):
LGTM !

builder/common/run_config_test.go
kp2099: Added the message.

anshulsharma-hashicorp (approved):
LGTM

PR #483, opened 2024-05-23
common: validate capacity reservation options

### LINKED ISSUE #473: Error Encountered When Building AMI with capacity_reservation_id
#### Error Encountered When Building AMI with capacity_reservation_id

I'm currently attempting to utilize the capacity_reservation_id attribute in Packer to launch the template. However, I keep encountering the following error:
**_Build 'amazon-ebs.autogenerated_1' errored after 1 second 875 milliseconds: Error launching source instance: InvalidParameterCombination: The request can include either a CapacityReservationId or a CapacityReservationPreference but not both._**

##### Reproduction Steps
Steps to reproduce this issue

##### Packer version
1.10.2

##### Simplified Packer Template
```
packer {
  required_version = ">= 1.10.2"
  required_plugins {
    amazon = {
      version = ">= 1.3.1"
    }
    ansible = {
      source  = "github.com/hashicorp/ansible"
      version = "~> 1"
    }
  }

source "amazon-ebs" "autogenerated_1" {
  ami_description = "Image for P5."
  ami_name        = "Ami-p5test"
  max_retries = 1000
  capacity_reservation_id = "cr-0908086"
  iam_instance_profile    = "PackerBuildRole_profile"
  instance_type           = "p5.48xlarge"
  launch_block_device_mappings {
    delete_on_termination = true
    device_name           = "/dev/sda1"
    volume_size           = 300
    volume_type           = "gp2"
  }
  region       = "us-east-1"
  source_ami   = "${data.amazon-ami.autogenerated_1.id}"
  ssh_username = "ubuntu"
  subnet_id    = "subnet-04XXX"
  vpc_id = "vpc-XXX"
}
```

The problem lies in the code block between lines 890 and 892 of the following file: https://github.com/hashicorp/packer-plugin-amazon/blob/01cd50e05c90d905af09cc298b24011ec3227cd9/builder/common/run_config.go. Ideally, we shouldn't set it to "none" because doing so disables the use of the capacity_reservation_id. According to AWS, we can only pass one parameter in this scenario. Ref: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CapacityReservationSpecification.html

> Describes an instance's Capacity Reservation targeting option. You can specify only one parameter at a time. If you specify CapacityReservationPreference and CapacityReservationTarget, the request fails

Any help would be much appreciated.

Thank you!

lbajolet-hashicorp:
The capacity reservation options (ID, ARN, or Preference) are all mutually exclusive, only one of the three should be set in a given configuration.

The code wasn't enforcing this, leading to cases in which conflicts could be specified in the configuration, and the call to start the instance would then fail during build instead of before it starts.

This commit adds some logic to ensure that only valid options are accepted, and some tests are added to make sure these options cannot conflict.

Closes #473

### DIFF
--- a/builder/common/run_config.go
+++ b/builder/common/run_config.go
@@ -887,11 +887,24 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 
 	}
 
-	if c.CapacityReservationPreference == "" {
+	capacityReservationTargetSet := false
+	if c.CapacityReservationId != "" || c.CapacityReservationGroupArn != "" {
+		capacityReservationTargetSet = true
+	}
+
+	if c.CapacityReservationGroupArn != "" && c.CapacityReservationId != "" {
+		errs = append(errs, fmt.Errorf("capacity_reservation_id and capacity_reservation_group_arn are mutually exclusive, only one should be used"))
+	}
+
+	if capacityReservationTargetSet && c.CapacityReservationPreference != "" {
+		errs = append(errs, fmt.Errorf("capacity_reservation_id, capacity_reservation_group_arn and capacity_reservation_preference are mutually exclusive, only one should be set"))
+	}
+
+	if c.CapacityReservationPreference == "" && c.CapacityReservationId == "" && c.CapacityReservationGroupArn == "" {
 		c.CapacityReservationPreference = "none"
 	}
 	switch c.CapacityReservationPreference {
-	case "none", "open":
+	case "", "none", "open":
 	default:
 		errs = append(errs, fmt.Errorf(`capacity_reservation_preference only accepts 'none' or 'open' values`))
 	}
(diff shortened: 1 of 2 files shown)


builder/common/run_config.go
```diff
@@ -887,11 +887,24 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
...
 	}
 
-	if c.CapacityReservationPreference == "" {
+	capacityReservationTargetSet := false
+	if c.CapacityReservationId != "" || c.CapacityReservationGroupArn != "" {
+		capacityReservationTargetSet = true
+	}
+
+	if c.CapacityReservationGroupArn != "" && c.CapacityReservationId != "" {
+		errs = append(errs, fmt.Errorf("capacity_reservation_id and capacity_reservation_group_arn are mutually exclusive, one only should be used"))
```
JenGoldstrich, line 896: Nit: I think `only one` reads a bit cleaner 

```suggestion
		errs = append(errs, fmt.Errorf("capacity_reservation_id and capacity_reservation_group_arn are mutually exclusive, only one should be used"))
```

JenGoldstrich (approved):

PR #490, opened 2024-06-21
builder: remove instances of confighelper.Trilean

lbajolet-hashicorp:
Since most of those instances of confighelper were not necessary as there is not a config package, nor is there local name conflicts as well, we can remove those instaces, especially the one from Config as it caused the (fragile) typechecks from the SDK to not render it as a boolean in the web docs otherwise.

Related to https://github.com/hashicorp/packer-plugin-sdk/pull/244


sylviamoss (approved):
👍🏼

Issue #492, opened 2024-07-05
Can not provide arn for organization


hc-github-team-packer:
_This issue was originally opened by @EugenKon in https://github.com/hashicorp/packer/issues/13099 and has been migrated to this repository. The original issue description is below._

---
#### Overview of the Issue

![image](https://github.com/hashicorp/packer/assets/139780495/670bb730-05c7-4aee-a002-1cefa0e9278e)

also this is not documented https://developer.hashicorp.com/packer/integrations/hashicorp/amazon/latest/components/builder/ebs how to fill this value correctly.

#### Reproduction Steps
```
source "amazon-ebs" "nomad_ami" {
  ami_name              = "xxx"
  instance_type         = "t2.medium"
  region                = var.aws_region
  source_ami            = data.amazon-ami.ubuntu_server_lts.id
  ssh_username          = "ubuntu"
  ami_regions           = ["ca-central-1", "us-east-1", "us-east-2", "us-west-1", "us-west-2"]
  ami_org_arns          = ["arn:aws:organizations::000000000:ou/path/to"]
  force_deregister      = true
  force_delete_snapshot = true

  tags = {
    Name          = "nomad"
    Base_AMI_ID   = data.amazon-ami.ubuntu_server_lts.id
    Base_AMI_Name = data.amazon-ami.ubuntu_server_lts.name
    OS_Version    = "Ubuntu"
    Release       = "24.04 LTS"
    Description   = "Base image for EC2 instances on Nomad cluster"
  }

  snapshot_tags = {
    Name        = "nomad"
    Description = "Base image for EC2 instances on Nomad cluster"
  }
}

```

### Packer version

v1.9.4

### Operating system and Environment details

Darwin Eugens-MacBook-Pro.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:09:52 PDT 2024; root:xnu-10063.121.3~5/RELEASE_X86_64 x86_64 i386 Darwin

EugenKon:
As a temporary workaround I commented out `ami_org_arns` option and applied ARN manually via AWS Console:
![image](https://github.com/hashicorp/packer-plugin-amazon/assets/139780495/6993d6b6-4520-484f-86d3-685d17d3bcbc)

lbajolet-hashicorp:
Hi @EugenKon,

Thanks for reporting this, regarding the error you're getting, that's from the AWS APIs it seems, but it's hard to understand the root cause of the issue here without more information.

[...]

PR #494, opened 2024-07-16
builder: check Org/OU ARNs during prepare

lbajolet-hashicorp:
When a builder accepts organisation or organisational unit ARNs as part of its config, we don't actually validate that the format matches what AWS accepts, leading to confusion among our users.

This commit checks that the provided ARNs match the expected format before running the build, and tries to provide information on what to change in the configs so they succeed later.

Related to #492


builder/common/ami_config_test.go
```diff
@@ -283,3 +283,78 @@ func TestAMINameValidation(t *testing.T) {
...
 
 }
+
+// Make sure only valid ARNs are accepted
+func TestAMIARNValidation(t *testing.T) {
+	tests := []struct {
+		name      string
+		OrgARN    []string
+		OUARN     []string
+		expectErr bool
```
JenGoldstrich, line 293: I would consider extending this to check the specific error returned, since there's several different cases that could change and these tests wouldn't fail

builder/common/ami_config.go
```diff
@@ -195,6 +201,34 @@ func (c *AMIConfig) Prepare(accessConfig *AccessConfig, ctx *interpolate.Context
...
+		if !OUARNRegexp.MatchString(ouARN) {
+			if OrgARNRegexp.MatchString(ouARN) {
+				errs = append(errs,
+					fmt.Errorf("Organisation Unit ARN %q looks like a Organisation ARN, "+
+						"it should be part of the `ami_org_arns` collection instead", ouARN))
+				continue
+			}
+
+			errs = append(errs, fmt.Errorf("Invalid Organisation Unit ARN %q, expect something that matches the "+
+				"following: ^arn:aws:organizations::\\d{12}:ou\\/o-[a-z0-9]{10,32}\\/ou-[0-9a-z]{4,32}-[0-9a-z]{8,32}", ouARN))
```
JenGoldstrich, line 228: I think for this error it may be helpful to include an example of the valid ARN for each case, as regular expressions can be challenging for novices to quickly parse.

JenGoldstrich (approved):
LGTM largely, and is definitely an improvement from the current experience, nice work, I made two suggestions, feel free to take them or leave them

builder/common/ami_config.go
lbajolet-hashicorp: Yeah, I imagine we can provide an example instead of a regexp for both; I'll amend that

builder/common/ami_config_test.go
lbajolet-hashicorp: I'm less sure about this, I think checking if the configs provoke an error when expected is more important than the contents of those errors, so I'm tempted to leave it as-is, but please feel free to push back if you think it's important

JenGoldstrich (approved):

PR #499, opened 2024-08-08
deps: bump go-retryablehttp to v0.7.7

dduzgun-security:
Bump go-retryablehttp to v0.7.7 to resolve https://github.com/hashicorp/packer-plugin-amazon/security/dependabot/30


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hashicorp/packer-plugin-amazon
 
-go 1.21.0
+go 1.21.13
```
lbajolet-hashicorp, line 3: Is there a reason to bump this? `.go-version` changes which version of go is used to build the plugin binaries, so by bumping this, the next releases will be built with this version, as for the `go` directive here, I'd be tempted to leave it at `1.21.0` since it's the minimum version you have to use to build the plugin as of today.

lbajolet-hashicorp (commented):
Hi @dduzgun-security,

Thanks for the PR! I left a comment on bumping the minimum go version, I would suggest we don't to avoid anyone developing the plugin having to install a more recent version of the go toolchain for local development.

On another note, for curiosity purposes, why are we bumping the go-version here? I understand that `retryablehttp` gets bumped to address the linked CVE, but it seems that they are still requiring go 1.19 in their modfile, so this looks unrelated; not against the idea of building the plugin with a newer version to be clear, just want to understand the rationale here.

Other than that, LGTM!

lbajolet-hashicorp (approved):
LGTM!

go.mod
dduzgun-security: @lbajolet-hashicorp thanks a lot for the review. 🙌  I removed this part for now as we discussed internally to tackle go bumps in a second PR

dduzgun-security:
@lbajolet-hashicorp it seems like I don't have the GitHub permissions to merge. 😄 
Would you be able to click on `Squash and merge` for me please?

lbajolet-hashicorp:
Yessir, I was waiting for the tests to go green, doing this now :)

PR #501, opened 2024-08-12
feat: enable setting a snapshot description for EBS surrogate

gnought:
Similar to EBS volume, this PR extends the support of setting Snapshot description in EBS surrogate

### DIFF
--- a/builder/ebssurrogate/step_snapshot_volumes.go
+++ b/builder/ebssurrogate/step_snapshot_volumes.go
@@ -24,13 +24,14 @@ import (
 //
 //	snapshot_ids map[string]string - IDs of the created snapshots
 type StepSnapshotVolumes struct {
-	PollingConfig   *awscommon.AWSPollingConfig
-	LaunchDevices   []*ec2.BlockDeviceMapping
-	snapshotIds     map[string]string
-	snapshotMutex   sync.Mutex
-	SnapshotOmitMap map[string]bool
-	SnapshotTags    map[string]string
-	Ctx             interpolate.Context
+	PollingConfig       *awscommon.AWSPollingConfig
+	LaunchDevices       []*ec2.BlockDeviceMapping
+	snapshotIds         map[string]string
+	snapshotMutex       sync.Mutex
+	SnapshotOmitMap     map[string]bool
+	SnapshotTags        map[string]string
+	SnapshotDescription string
+	Ctx                 interpolate.Context
 }
 
 func (s *StepSnapshotVolumes) snapshotVolume(ctx context.Context, deviceName string, state multistep.StateBag) error {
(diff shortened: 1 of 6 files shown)


builder/ebssurrogate/step_snapshot_volumes.go
```diff
@@ -68,13 +68,15 @@ func (s *StepSnapshotVolumes) snapshotVolume(ctx context.Context, deviceName str
 			ResourceType: aws.String("snapshot"),
 			Tags:         snapshotTags,
 		}
-
 		tagSpecs = append(tagSpecs, snapTags)
 	}
-
+	description := s.SnapshotDescription
+	if description == "" {
+		description = fmt.Sprintf("Packer: %s", time.Now().String())
```
lbajolet-hashicorp, line 75: I would suggest leaving the description empty be default, if only not to change the current behaviour by default, but also as the proposed description does not add information to the snapshot itself.

builder/ebsvolume/step_snapshot_ebs_volumes.go
```diff
@@ -55,11 +56,14 @@ func (s *stepSnapshotEBSVolumes) Run(ctx context.Context, state multistep.StateB
 					ResourceType: aws.String("snapshot"),
 					Tags:         tags,
 				}
-
+				description := configVolumeMapping.SnapshotDescription
+				if description == "" {
+					description = fmt.Sprintf("Packer: %s", time.Now().String())
```
lbajolet-hashicorp, line 61: Same here, I would suggest leaving the description empty by default.

lbajolet-hashicorp (commented):
Hi @gnought,

Thanks for the PR!
I left a small comment on the default value, I think the empty default is good enough that we don't need to generate one if not provided.

Aside from that nit, LGTM! Pre-approving so it is not a blocker later.

lbajolet-hashicorp (approved):

lbajolet-hashicorp:
Note: I see you haven't signed the CLAs, would you be able to do so? If not we won't be able to accept the contribution for legal reasons, sorry about that.

builder/ebssurrogate/step_snapshot_volumes.go
gnought: Thanks for the comment.
The fix is pushed.

builder/ebsvolume/step_snapshot_ebs_volumes.go
gnought: Thanks for the comment.
The fix is pushed.

gnought:
signed CLA and fixes are pushed. Please review. :)

lbajolet-hashicorp:
Hi @gnought,

I see I misunderstood something in my previous review: the description you added was already the default (which was not changeable in the current state), so we should continue forwarding it, apologies for missing that.
The PR in its previous state was good, no need for the rerolls; given they're separate commits, I'll force-push this branch again without them, and merge as soon as tests go green.

Thanks for the update!

Issue #506, opened 2024-08-30
Enable Fast Launch by Region


JasonLibeyTR:
#### Community Note

Please vote on this issue by adding a 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to the original issue to help the community and maintainers prioritize this request.
Please do not leave "+1" or "me too" comments, they generate extra noise for issue followers and do not help prioritize the request.
If you are interested in working on this issue or have submitted a pull request, please leave a comment.

#### Description

The current way Fast Launch is supported assumes you want to enable it on all the regions you are copying the AMI to. The enable_fast_launch configuration is only at the root level of the Fast Launch configuration. There should be a way to disable Fast Launch for some regions.

#### Use Case(s)

An AMI that is baked is copied to multiple regions. Only one region sees heavy use and will benefit from having Fast Launch enabled. The other regions should not have to be enabled as that is a waste of resources/money.

#### Potential configuration

Add the enable_fast_launch boolean to the region_launch_templates array.
```
      "fast_launch": {
        "enable_fast_launch": "true",
        "region_launch_templates": [
          {"region": "us-east-1", "template_id": "lt-04eb90cb108a19b0b", "template_version": 6 },
          {"region": "us-west-2", "enable_fast_launch": "false" }
        ],
        "max_parallel_launches": 6,
        "target_resource_count": 5
      }
```

lbajolet-hashicorp:
Hi @JasonLibeyTR,

Thanks for reporting this! I've just opened a PR that introduces `enable_fast_launch` to the `region_launch_templates` array, this will be rolled into an upcoming release of the plugin, but in the meantime (ideally before we merge it), may I ask you if you can test the plugin with this addition?

You'll need to compile and install it to your local environment, feel free to refer to our [docs](https://developer.hashicorp.com/packer/docs/plugins/install#manually-install-plugins-using-the-cli) for information on that, alternatively if you have `make` installed, you can invoke `make dev` which will build and install the plugin so you can run tests with it.

PR #509, opened 2024-09-16
Allow fast boot disable on region

lbajolet-hashicorp:
The fast-launch configuration allows setupping fast-launch for Windows virtual machines, either with a common configuration, or with a region-by-region configuration.
However if someone wants to only enable fast-launch on a subset of regions, there wasn't any way to express that, leading to either a waste of resources, or a need to manually cleanup the extra snapshots created.

Therefore this commit adds an extra trilean to the regional configuration for fast-launch, so it can be explicitly disabled on selected regions.

Closes: #506

### DIFF
--- a/builder/ebs/fast_launch_setup.go
+++ b/builder/ebs/fast_launch_setup.go
@@ -9,6 +9,8 @@ package ebs
 import (
 	"fmt"
 	"log"
+
+	"github.com/hashicorp/packer-plugin-sdk/template/config"
 )
 
 // FastLaunchTemplateConfig is the launch template configuration for a region.
@@ -17,6 +19,18 @@ import (
 // in the template, as each fast-launch enablement step occurs after the
 // copy, and each region may pick their own launch template.
 type FastLaunchTemplateConfig struct {
+	// Enable fast launch allows you to disable fast launch settings on the region level.
+	//
+	// If unset, the default region behavior will be assumed - i.e. either use
+	// the globally specified template ID/name (if specified), or AWS will set
+	// it for you.
+	//
+	// Using other fast launch options, while unset, will imply enable_fast_launch to be true.
+	//
+	// If this is explicitly set to `false` fast-launch will be
+	// disabled for the specified region and all other options besides region
+	// will be ignored.
+	EnableFalseLaunch config.Trilean `mapstructure:"enable_fast_launch"`
 	// The region in which to find the launch template to use
 	Region string `mapstructure:"region" required:"true"`
 	// The ID of the launch template to use for the fast launch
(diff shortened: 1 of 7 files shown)


builder/ebs/fast_launch_setup.go
```diff
@@ -17,6 +19,16 @@ import (
...
 type FastLaunchTemplateConfig struct {
+	// EnableFalseLaunch allows you to specify if fast launch should be
+	// enabled on a region-by-region basis
+	//
+	// If unset, this will assume the default behaviour, i.e. the other
+	// options will imply that this is true.
+	//
+	// If this is explicitly set to `false` however, fast-launch will be
+	// disabled for the specified region, and options other than the region
+	// will be ignored.
```
nywilken, line 30: Small rewrite to fix typo and help clarify the intent. 
Is the default behavior always true or based on some condition? 

```suggestion
	// Enable fast launch allows you to disable fast launch settings on the region level.
	//
	// If unset, the default region behavior will be assumed.
	// Using other fast launch options, while unset, will imply enable_fast_launch to be true.
	//
	// If this is explicitly set to `false` fast-launch will be
	// disabled for the specified region and all other options besides region
	// will be ignored.
```

builder/ebs/step_prepare_fast_launch_template.go
```diff
@@ -49,6 +54,14 @@ func (s *stepPrepareFastLaunchTemplate) Run(ctx context.Context, state multistep
 	for _, templateSpec := range s.RegionTemplates {
 		region := templateSpec.Region
 
+		if templateSpec.EnableFalseLaunch == config.TriFalse {
+			log.Printf("[INFO] fast-launch explicitly disabled for region %q", region)
+			templateIDsByRegion[region] = TemplateSpec{
+				Enabled: false,
```
nywilken, line 60: I'm not to clear on how this works by default but from what I gather (looking at the issue) fast launch seems to be enabled by default on the region. Is that correct? 

If so, does the user need any particular permissions to disable on the region?

nywilken (commented):
Overall this looks good. I left a suggestion on the documentation and a question about permissions, if needed.

builder/ebs/fast_launch_setup.go
lbajolet-hashicorp: Yeah so the default behaviour is implicitly true if there's a `region_launch_templates` defined with at least one of the mandatory attributes besides `region` (refer to the [`Prepare`](https://github.com/hashicorp/packer-plugin-amazon/blob/main/builder/ebs/fast_launch_setup.go#L43) implementation for the structure for more info).
If explicitely set to false, this just disables fast-launch for the region, as is requested in the linked issue.

builder/ebs/step_prepare_fast_launch_template.go
lbajolet-hashicorp: No permissions necessary, we enable fast-launch for each region in a later step, so if you don't want it on a particular region, we don't request it be enabled

nywilken (approved):
Thanks for the clarification. LGTM

PR #518, opened 2024-12-05
Fix 2022 sysprep provisioner cmd

drewmullen:
The sysprep commands currently in the documentation do not work anymore. That is because they are written for EC2Launch v1. V2 moved to new executables in a different location where previously v1 used scripts. v2 is the default for all windows AMIs now as far as i can tell.

The new addition to the docs I have confirmed work for 2022. However, I was not able to get a 2016 or 2019 server working with sysprep provisioners commands. I'm leaving them as is

### DIFF
--- a/.web-docs/components/builder/ebs/README.md
+++ b/.web-docs/components/builder/ebs/README.md
@@ -2215,6 +2215,34 @@ build {
 }
 ```
 
+## Windows 2022 Sysprep Commands - For Amazon Windows AMIs Only
+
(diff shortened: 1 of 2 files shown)


lbajolet-hashicorp (approved):
Hey @drewmullen,

Thanks for that PR, the amendment looks good to me; merging now!

lbajolet-hashicorp (approved):

JenGoldstrich (approved):

PR #523, opened 2025-01-07
Add run tags to amazon-ebsvolume IAM role

### LINKED ISSUE #522: amazon-ebsvolume run_tags are not applied to the IAM profile

#### Overview of the Issue

`run_tags` are not applied to the IAM role when using `amazon-ebsvolume`. This appears similar to https://github.com/hashicorp/packer-plugin-amazon/issues/322

#### Reproduction Steps

Specify `run_tags` when building using `amazon-ebsvolume`. They will not get applied to the temporary IAM role.

### Plugin and Packer version

From `1.3.4`

All

N/A - no crash or other visible error

OompahLoompah:
When using the amazon-ebsvolume builder ensure that run_tags are applied to the IAM profile, consistent with the amazon-ebs builder.

Closes #522

### DIFF
--- a/builder/ebsvolume/builder.go
+++ b/builder/ebsvolume/builder.go
@@ -316,6 +316,8 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			IamInstanceProfile:                        b.config.IamInstanceProfile,
 			SkipProfileValidation:                     b.config.SkipProfileValidation,
 			TemporaryIamInstanceProfilePolicyDocument: b.config.TemporaryIamInstanceProfilePolicyDocument,
+			Tags: b.config.RunTags,
+			Ctx:  b.config.ctx,
 		},
 		instanceStep,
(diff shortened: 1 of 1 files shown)


kp2099 (commented):
Hi @OompahLoompah, thank you for your contribution!

The change looks good to me, and the implementation makes sense. I believe we’ll be in a good position to merge this soon.

One thing I noticed — it looks like the tags aren't currently being propagated to the IAM instance profile in other builders, such as `instance` and `ebs-surrogate`. Would you be open to extending this change to cover those builders as well? If not, no worries — I’d be happy to follow up with a separate PR.

As for testing, it seems that the current acceptance test coverage is mostly focused on the post-build steps. I think we can handle any additional testing improvements as a follow-up.

Let me know what you think. I appreciate your work on this!

kp2099:
Hi @OompahLoompah, since these changes are now included in #553, I'm closing this pull request to avoid duplication.

PR #533, opened 2025-03-21
Speed up and fix acceptance tests

mogrogan:
- **ebssurrogate: use t.Parallel and fix acc tests**
- **datasource: use t.Parallel and fix acc tests**
- **ebs: use t.Parallel and fix acc tests**

Closes #xxx
Closes #xxx


.github/workflows/acceptance-tests.yml
```diff
@@ -61,8 +67,8 @@ jobs:
           role-duration-seconds: 21600 # 6h
       - run: |
           echo "Testing with Go ${{ needs.get-go-version.outputs.go-version }}"
-          PACKER_ACC=1 go test \
```
JenGoldstrich, line 64: I believe PACKER_ACC=1 is required here, otherwise acceptance tests will not run

lbajolet-hashicorp: Since it's set from the environment above you can remove it from the command-line here, I think this is OK

.github/workflows/acceptance-tests.yml
```diff
@@ -61,8 +67,8 @@ jobs:
...
-            -timeout 5h \
+          go test \
+            -timeout 2h \
```
lbajolet-hashicorp, line 71: While I would assume they don't need 5 hours to run, I think it might be premature to lower that to 2h, it's a timeout, we can be large here, especially since if we're too optimistic this will `kill -9` the tests, and not clean-up the resources.
All that to say, I think we shouldn't change that now, I think it doesn't hurt to be pessimistic with the timeout here.

builder/ebs/builder_acc_test.go
```diff
@@ -1081,8 +1105,10 @@ func TestAccBuilder_EbsWindowsFastLaunch(t *testing.T) {
 		},
 	}
 
-	for _, tt := range tests {
+	for _, tc := range tests {
+		tt := tc
```
lbajolet-hashicorp, line 1109: If we alias `tc` to `tt` in the test to minimise the amount of changes, I would argue we can keep the current name for those local test variables

builder/ebs/builder_acc_test.go
```diff
@@ -1425,13 +1453,15 @@ func TestAccBuilder_AssociatePublicIPWithoutSubnet(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			tc := tt
```
lbajolet-hashicorp, line 1456: I would suggest to pick one and stay consistent, if we have `tt` all around we can keep that, and later on we can do a pass if we think `tc` is the better name for those

.github/workflows/acceptance-tests.yml
JenGoldstrich: Oh derp, my bad.

builder/ebs/builder_acc_test.go
```diff
@@ -2218,6 +2253,7 @@ build {
 	sources = ["amazon-ebs.windows-fastboot"]
 
 	provisioner "powershell" {
+		skip_clean = true
```
lbajolet-hashicorp, line 2256: Is this necessary? I know we discussed it out-of-band yesterday, but IIRC we didn't conclude this was the reason why the tests failed; if it is the reason though, I think it warrants a documentation update to highlight that it should be set for sysprep when using fast-boot.

lbajolet-hashicorp (commented):
Overall LGTM, I left a few comments on some of the changes, but they're reasonably minor.

Thanks for taking care of fixing those @mogrogan, they needed some love!

builder/ebs/builder_acc_test.go
mogrogan: I chosed to take currTest as it was already in used for a different test and make it more clear the variable purpose is to capture context

lbajolet-hashicorp (approved):
Largely LGTM, there's still the consistency issue on some EBS acctests, but this can be addressed in a later PR. Merging this one now

PR #537, opened 2025-04-08
ci: use the proper datasource target for acc tests

mogrogan:


lbajolet-hashicorp (approved):
Ah yeah good catch, sorry I missed it in my previous reviews; LGTM!

Issue #361, opened 2023-04-12
Provide way of unsetting run tags on AMI


frh-td:
We have some scripts running on our instances that we don't want to run on the instances being used to create AMIs. We've been detecting if the instance is a packer-launched instance by setting a specific instance tag in `run_tags`, then checking for the presence of this tag in those scripts, and aborting if present.

With https://github.com/hashicorp/packer-plugin-amazon/pull/327 those tags end up also being set on the created AMIs, which is undesirable. I tried to see if I could unset a tag from `run_tags` by setting it to `null` in `tags`, but it aborted saying "Error: null value is not allowed".

Is it possible to allow unsetting a run tag on the created AMI, or is there perhaps some alternative way for a process to determine if the instance it is running on is a Packer-created instance? Alternative solutions we have come up with so far are to use the value of a tag that we set differently in `run_tags` and `tags` (like `Name`), or to pass a value through user data.

PR #538, opened 2025-04-21
 Skip attaching run tags to the finished AMI.

kp2099:
Provide a way to users to allow them to skip the tags that are created to be attached to the AMI.
As of today we propagate the tags that are set in run_tags to the AMI. I propose on adding a new config

**_skip_ami_run_tags_** : Setting this config to True would make sure the tags set in run_tags are not propagated to the finished AMI instance.

Closes: https://github.com/hashicorp/packer-plugin-amazon/issues/361

### DIFF
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -722,6 +722,42 @@ func TestAccBuilder_EbsRunTagsJSON(t *testing.T) {
 	acctest.TestPlugin(t, testcase)
 }
 
+//go:embed test-fixtures/interpolated_skip_run_tags.pkr.hcl
+var testInterpolatedSkipRunTagsSource string
+
+func TestAccBuilder_EbsSkipAmiRunTags(t *testing.T) {
+	//t.Parallel()
+	ami := amazon_acc.AMIHelper{
+		Region: "us-west-2",
+		Name:   fmt.Sprintf("packer-amazon-skip-ami-run-tags-test %d", time.Now().Unix()),
+	}
+
+	testcase := &acctest.PluginTestCase{
+		Name: "amazon-ebs_skip_ami_run_tags_test",
+		Teardown: func() error {
+			return ami.CleanUpAmi()
+		},
+		Template: fmt.Sprintf(testInterpolatedSkipRunTagsSource, ami.Name),
+		Check: func(buildcommand *exec.Cmd, logfile string) error {
+			var result error
+			if buildcommand.ProcessState != nil {
+				if buildcommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("bad exit code. logfile: %s", logfile)
+				}
+			}
+			// empty tag map since we should not attach any tags if skip_ami_run_tags is set
+			expectedTags := map[string]string{}
+			err := checkAMITags(ami, expectedTags)
+			if err != nil {
+				result = multierror.Append(result, err)
+			}
+
+			return result
+		},
+	}
+	acctest.TestPlugin(t, testcase)
+}
+
 //go:embed test-fixtures/ssh-keys/rsa_ssh_keypair.pkr.hcl
 var testSSHKeyPairRSA string
 
@@ -1360,6 +1396,11 @@ func checkAMITags(ami amazon_acc.AMIHelper, tagList map[string]string) error {
 
 	var errs error
 	image := imageResp.Images[0] // Only requested a single AMI ID
+	if len(tagList) == 0 {
+		if len(image.Tags) != 0 {
+			return fmt.Errorf("expected no tags for AMI %q, got %d", amiNameRegion, len(image.Tags))
+		}
+	}
 	for tagKey, tagVal := range tagList {
 		found := false
 		for _, imgTag := range image.Tags {
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -726,7 +726,7 @@ func TestAccBuilder_EbsRunTagsJSON(t *testing.T) {
 var testInterpolatedSkipRunTagsSource string
 
 func TestAccBuilder_EbsSkipAmiRunTags(t *testing.T) {
-	//t.Parallel()
+	t.Parallel()
 	ami := amazon_acc.AMIHelper{
 		Region: "us-east-1",
 		Name:   fmt.Sprintf("packer-amazon-skip-ami-run-tags-test %d", time.Now().Unix()),
--- a/builder/ebs/builder_acc_test.go
+++ b/builder/ebs/builder_acc_test.go
@@ -728,7 +728,7 @@ var testInterpolatedSkipRunTagsSource string
 func TestAccBuilder_EbsSkipAmiRunTags(t *testing.T) {
 	//t.Parallel()
 	ami := amazon_acc.AMIHelper{
-		Region: "us-west-2",
+		Region: "us-east-1",
 		Name:   fmt.Sprintf("packer-amazon-skip-ami-run-tags-test %d", time.Now().Unix()),
 	}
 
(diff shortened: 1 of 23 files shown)


builder/ebs/step_create_ami.go
```diff
@@ -74,7 +75,12 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 			return multistep.ActionHalt
 		}
 
-		createOpts.TagSpecifications = ec2Tags.TagSpecifications(ec2.ResourceTypeImage, ec2.ResourceTypeSnapshot)
+		if !s.AMISkipRunTags {
+			ui.Say(fmt.Sprintf("Attaching run tags to AMI..."))
+			createOpts.TagSpecifications = ec2Tags.TagSpecifications(ec2.ResourceTypeImage, ec2.ResourceTypeSnapshot)
+		} else {
+			ui.Say(fmt.Sprintf("Skipping attaching run tags to AMI..."))
+		}
```
anshulsharma-hashicorp, line 83: Please look at the suggestion from the github AI agent, if you are not formatting string you can avoid use of fmt.Sprintf method.

mogrogan (commented):
I think it's look good, consider adding new acceptance tests for these

builder/ebssurrogate/builder_acc_test.go
```diff
@@ -282,6 +359,38 @@ build {
...
+	}
+	ami_virtualization_type = "hvm"
+	ami_root_device {
+		source_device_name = "/dev/xvda"
+		device_name = "/dev/sda1"
+		delete_on_termination = true
+		volume_size = 8
+		volume_type = "gp2"
+	}
+	run_tags = {
```
mogrogan, line 383: what if I decided to add tags within the `tags` attributes https://developer.hashicorp.com/packer/integrations/hashicorp/amazon/latest/components/builder/ebssurrogate#ami-configuration

builder/ebssurrogate/step_create_ami.go
```diff
@@ -73,6 +74,12 @@ func (s *StepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 			ui.Error(err.Error())
 			return multistep.ActionHalt
 		}
+		if !s.AMISkipRunTags {
```
mogrogan, line 77: same comment as in ebs I guess

mogrogan (changes requested):
Thank you for your contribution :) 

Like I said in the comments I am not sure what would happens if we decided to add `tags` and no `run_tags` I would start by adding a test to check that out and fix the issue if there is.

But solid work nevertheless

builder/ebssurrogate/step_create_ami.go
```diff
@@ -73,6 +74,12 @@ func (s *StepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
...
+			ui.Say("Attaching run tags to AMI...")
+			createOpts.TagSpecifications = ec2Tags.TagSpecifications(ec2.ResourceTypeImage, ec2.ResourceTypeSnapshot)
+		} else {
+			ui.Say("Skipping attaching run tags to AMI...")
+		}
 
 		createOpts.TagSpecifications = ec2Tags.TagSpecifications(ec2.ResourceTypeImage, ec2.ResourceTypeSnapshot)
```
anshulsharma-hashicorp, line 84: @kp2099 i think you forgot to remove this, this is already taken care in the if part.

anshulsharma-hashicorp: with this, did the acceptance test pass, i think this is something we can recheck, if test passes even with this line  then this is not correct behavior.

builder/ebssurrogate/builder_acc_test.go
```diff
@@ -51,6 +97,37 @@ func TestAccBuilder_EbssurrogateBasic(t *testing.T) {
...
+		Check: func(buildCommand *exec.Cmd, logfile string) error {
+			var result error
+			if buildCommand.ProcessState != nil {
+				if buildCommand.ProcessState.ExitCode() != 0 {
+					return fmt.Errorf("Bad exit code. Logfile: %s", logfile)
+				}
+			}
+
+			expectedTags := map[string]string{}
+			err := checkAMITags(ami, expectedTags)
```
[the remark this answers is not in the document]
kp2099: Addressed in the above comment, on why the acceptance tests worked

builder/ebs/test-fixtures/interpolated_skip_run_tags.pkr.hcl
```diff
@@ -0,0 +1,42 @@
+# Copyright (c) HashiCorp, Inc.
```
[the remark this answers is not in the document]
kp2099: Yes @mogrogan i believe maintaining test files pkr templates this way is much cleaner than inline strings. I have refactored the tests in the ebs surrogate to use a test-fixture structure and individual pkr files that gets embedded to the test suite.

builder/ebs/test-fixtures/interpolated_skip_run_tags.pkr.hcl
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: MPL-2.0
+
+data "amazon-ami" "test" {
```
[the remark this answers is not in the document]
kp2099: good idea. I have directly used the source_ami. Removed the data block

builder/ebs/builder_acc_test.go
```diff
@@ -722,6 +722,42 @@ func TestAccBuilder_EbsRunTagsJSON(t *testing.T) {
 	acctest.TestPlugin(t, testcase)
 }
 
+//go:embed test-fixtures/interpolated_skip_run_tags.pkr.hcl
+var testInterpolatedSkipRunTagsSource string
+
+func TestAccBuilder_EbsSkipAmiRunTags(t *testing.T) {
```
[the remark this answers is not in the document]
kp2099: it wont be wiped. if we see the steps in the ebs and ebs surrogate builder, after creating the AMI, we have a dedicated step to attach the tags that are set using the "tag" config

reference: https://github.com/hashicorp/packer-plugin-amazon/blob/edc76f2d4fa0a671cee5ccbd591754c71c43388c/builder/ebs/builder.go#L467-L472

Nevertheless, i added a test case to demonstrate that the `tags` set in a packer file do get set irrespective of the `skip_ami_run_tags` config that we are introducing

builder/ebs/step_create_ami.go
```diff
@@ -74,7 +75,12 @@ func (s *stepCreateAMI) Run(ctx context.Context, state multistep.StateBag) multi
 			return multistep.ActionHalt
 		}
 
-		createOpts.TagSpecifications = ec2Tags.TagSpecifications(ec2.ResourceTypeImage, ec2.ResourceTypeSnapshot)
+		if !s.AMISkipRunTags {
+			ui.Say("Attaching run tags to AMI...")
+			createOpts.TagSpecifications = ec2Tags.TagSpecifications(ec2.ResourceTypeImage, ec2.ResourceTypeSnapshot)
```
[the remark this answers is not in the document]
kp2099: as addressed above, the tags set using the "tags" config will be attached after the AMI is created.

builder/ebssurrogate/step_create_ami.go
kp2099: addressed above. added a new acceptance test to test the same

builder/ebssurrogate/step_create_ami.go
anshulsharma-hashicorp: Yes, i think this makes sense, thanks for taking care of this.

mogrogan (approved):
Really nice, LGTM

anshulsharma-hashicorp (dismissed):

builder/ebs/builder_acc_test.go
```diff
@@ -722,6 +722,80 @@ func TestAccBuilder_EbsRunTagsJSON(t *testing.T) {
...
+	//t.Parallel()
```
mogrogan, line 729: I think you forgot to uncomment this

mogrogan (changes requested):

builder/ebs/builder_acc_test.go
kp2099: oh sorry about that. removed it. thanks for catching it!

mogrogan (approved):

PR #544, opened 2025-05-19
Add IPv6-only support

siriusfreak:
### Summary

This merge request adds support for using `ipv6` as an `ssh_interface` in Amazon builders within Packer. This includes:

* Accepting `ipv6` as a valid `ssh_interface` value in the `RunConfig`.
* Resolving the IPv6 address of an EC2 instance when `ipv6` is specified.
* Adjusting the temporary security group logic to authorize ingress for IPv6 addresses by detecting CIDR format.
* Updating tests to cover the new `ipv6` interface.

### Motivation

Previously, Packer's EC2 builders did not support connecting over IPv6. This change allows users working in IPv6-preferred or IPv6-only environments to provision and communicate with EC2 instances using their IPv6 addresses.

### Implementation Details

* `RunConfig.Prepare`: added `ipv6` to the list of accepted `ssh_interface` values and defaulted security group source CIDR to `::/0` when `ipv6` is used.
* `SSHHost()`: added logic to extract the `Ipv6Address` field from EC2 instance metadata.
* `StepSecurityGroup.Run`: conditionally creates `IpRange` or `Ipv6Range` entries based on the presence of `:` in the CIDR block.
* Added `ipv6` support to the `fakeEC2Describer` in unit tests.
* Expanded the `TestSSHHost` test cases to verify `ipv6` behavior.

### Test Coverage

* ✅ Unit tests added for SSH interface resolution with `ipv6`.
* ✅ IPv6 address flow verified using `fakeEC2Describer`.

### Notes

This change is backward-compatible and does not alter behavior for existing SSH interfaces (`public_ip`, `private_ip`, etc.).

Please let me know if additional builder types should support `ipv6` or if further integration tests are desired.


kp2099 (changes requested):
Hi @siriusfreak, thank you for your contribution. I agree that, given the significant interest in IPv6 support, we should move forward with this change.

However, I noticed that some documentation updates will be necessary. Could you please add the relevant documentation changes as part of this PR?

ref: https://github.com/hashicorp/packer-plugin-amazon/blob/702cad3f102477821e1a24f4955d93221e01b7b4/.web-docs/components/builder/ebs/README.md#L1049

builder/common/ssh.go
```diff
@@ -57,6 +57,10 @@ func SSHHost(e ec2Describer, sshInterface string, host string) func(multistep.St
 					if i.PrivateDnsName != nil {
 						host = *i.PrivateDnsName
 					}
+				case "ipv6":
+					if i.Ipv6Address != nil {
+						host = *i.Ipv6Address
```
kp2099, line 62: Since the plugin uses the instance's `IPv6` address for SSH connections, the default VPC does not include an `IPv6` CIDR block, and its subnets typically don't have `"auto-assign IPv6 address"` enabled. As a result, instances launched in this default configuration won't receive an IPv6 address, and SSH will fail when `ssh_interface = "ipv6"` is used.

We should consider documenting that users need to launch instances in a VPC with an associated` IPv6` CIDR block and ensure the subnet has `"auto-assign IPv6 address"` enabled. This ensures that the instance gets an IPv6 address and that SSH connectivity will work as expected out of the box.

Let me know what you think.

builder/common/step_security_group.go
```diff
@@ -194,16 +200,31 @@ func (s *StepSecurityGroup) Run(ctx context.Context, state multistep.StateBag) m
...
-	groupRules := &ec2.AuthorizeSecurityGroupIngressInput{
-		GroupId: groupResp.GroupId,
-		IpPermissions: []*ec2.IpPermission{
-			{
-				FromPort:   aws.Int64(int64(port)),
-				ToPort:     aws.Int64(int64(port)),
-				IpRanges:   groupIpRanges,
-				IpProtocol: aws.String("tcp"),
+	var groupRules *ec2.AuthorizeSecurityGroupIngressInput
+	if len(groupIpv6Ranges) != 0 {
```
kp2099, line 204: i am trying to understand the need for the `if` check here, why do we need to assign a security group rule that is either `ipv6` or `ipv4`?
For example if a user provides a `temporary_security_group_cidrs` as 
```
temporary_security_group_source_cidrs = ["0.0.0.0/16", "51.207.193.182/32", "::/0"]
```
Wouldnt we want to apply both the ipv4 and ipv6 CIDRs to the security group.

i.e. 
```suggestion
groupRules = &ec2.AuthorizeSecurityGroupIngressInput{
	GroupId: groupResp.GroupId,
	IpPermissions: []*ec2.IpPermission{
		{
			FromPort:   aws.Int64(int64(port)),
			ToPort:     aws.Int64(int64(port)),
			Ipv6Ranges: groupIpv6Ranges,
			IpRanges:   groupIpRanges,
			IpProtocol: aws.String("tcp"),
		},
	},
}
```

siriusfreak:
I hope that I will be able to return soon

builder/common/run_config_test.go
```diff
@@ -592,3 +592,23 @@ func TestRunConfigPrepare_EnableSpotInstanceBadSpotAllocationStrategy(t *testing
...
+	}
+
+	if len(c.TemporarySGSourceCidrs) != 1 {
+		t.Fatalf("expected 1 default CIDR, got: %v", c.TemporarySGSourceCidrs)
+	}
+
+	if c.TemporarySGSourceCidrs[0] != "::/0" {
+		t.Errorf("expected default CIDR to be '::/0', got: %s", c.TemporarySGSourceCidrs[0])
+	}
+}
```
mukulmohan177, line 614: Hey instead of only using fake EC2 objects and mocks, why don't we add an integration test that actually launches an EC2 instance with IPv6 enabled in AWS ? We can then verify ssh connection using the new address. This would be better considering we can verify security groups.

builder/common/run_config.go
```diff
@@ -780,6 +784,7 @@ func (c *RunConfig) Prepare(ctx *interpolate.Context) []error {
 		c.SSHInterface != "private_ip" &&
 		c.SSHInterface != "public_dns" &&
 		c.SSHInterface != "private_dns" &&
+		c.SSHInterface != "ipv6" &&
 		c.SSHInterface != "session_manager" &&
 		c.SSHInterface != "" {
 		errs = append(errs, fmt.Errorf("Unknown interface type: %s", c.SSHInterface))
```
mukulmohan177, line 790: Hey can we add a clear error? Subnet / vpc does not support ipv6 
Also why don't we check if the user's env actually supports ipv6?

mukulmohan177 (changes requested):

builder/common/run_config.go
kp2099: Addressed internally

builder/common/run_config_test.go
kp2099: Addressed internally

builder/common/step_security_group.go
```diff
@@ -178,11 +178,17 @@ func (s *StepSecurityGroup) Run(ctx context.Context, state multistep.StateBag) m
 	// map the list of temporary security group CIDRs bundled with config to
 	// types expected by EC2.
 	groupIpRanges := []*ec2.IpRange{}
+	groupIpv6Ranges := []*ec2.Ipv6Range{}
 	for _, cidr := range temporarySGSourceCidrs {
-		ipRange := ec2.IpRange{
-			CidrIp: aws.String(cidr),
+		if strings.Contains(cidr, ":") {
```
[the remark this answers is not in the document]
kp2099: added net.ParseCIDR to segregate the CIDR ipv4 and ipv6 blocks

tanmay-hc (approved):

Issue #521, opened 2024-12-30
EBS volumes are left behind when using ebssurrogate with `use_create_image = true`


lorengordon:
#### Overview of the Issue

EBS volumes are not cleaned up when using ebssurrogate with `use_create_image = true`. From the logs below, both `vol-06a5bb097b689da36` and `vol-003b56d111f16eb56` remain behind, in an unattached "Available" state.

#### Reproduction Steps

Create an image using ebssurrogate and set `use_create_image = true`.

### Plugin and Packer version

```
❯ packer version
Packer v1.11.2

❯ packer plugins installed
/home/x/.config/packer/plugins/github.com/hashicorp/amazon/packer-plugin-amazon_v1.3.4_x5.0_linux_amd64
/home/x/.config/packer/plugins/github.com/hashicorp/azure/packer-plugin-azure_v2.1.7_x5.0_linux_amd64
/home/x/.config/packer/plugins/github.com/hashicorp/openstack/packer-plugin-openstack_v1.1.2_x5.0_linux_amd64
/home/x/.config/packer/plugins/github.com/hashicorp/vagrant/packer-plugin-vagrant_v1.1.4_x5.0_linux_amd64
/home/x/.config/packer/plugins/github.com/hashicorp/virtualbox/packer-plugin-virtualbox_v1.1.1_x5.0_linux_amd64
```

### Simplified Packer Buildfile

I don't have a _simple_ buildfile at the moment, hoping there's a test config already that can be executed to verify reproduction.

### Operating system and Environment details

Should be irrelevant...

### Log Fragments and crash.log files

```
==> amazon-ebssurrogate.minimal-ol-8-hvm: Detaching Root EBS Device Name: /dev/sda1, Volume ID: vol-06a5bb097b689da36
==> amazon-ebssurrogate.minimal-ol-8-hvm: Detaching Root EBS Device Name: /dev/xvdf, Volume ID: vol-003b56d111f16eb56
==> amazon-ebssurrogate.minimal-ol-8-hvm: Attaching Root EBS Device Name /dev/sda1, Volume ID: vol-003b56d111f16eb56
==> amazon-ebssurrogate.minimal-ol-8-hvm: Creating AMI spel-ci-minimal-ol-8-hvm-2024.12.dev1735591494.x86_64-gp3 from instance i-0f453ff06de2f64fe
    amazon-ebssurrogate.minimal-ol-8-hvm: AMI: ami-0e486a554fa252f78
[...]

PR #455, opened 2024-01-30
Add ability to EBS Surrogate builder to persist properties from Marketplace source image, such as product codes

dwc0011:
Add option to swap the root device to an EBS surrogate and create a new AMI in addition to being able to register an AMI.  This allows billing data to be retained when creating AMI's.  

Closes #440

PR #545, opened 2025-05-23
resetting the delete_on_termination attribute after reattaching the v…

kp2099:


kp2099:
> @kp2099 Are you using `use_create_image = true` by any chance? This could be related to #521? Though in that report, both volumes are getting left behind...

Hi @lorengordon — yes, I’ve been using use_create_image = true. From what I’ve observed, if any root volume is detached during the provisioning process and not reattached before instance termination, it won’t be deleted. For volumes to be automatically cleaned up, they need to remain attached to the instance when it’s terminated.

Is it possible that one of the volumes in the error log was detached and not reattached? That could explain why it wasn’t deleted. As for the volume that was reattached but still not deleted — this PR aims to address that specific issue.

kp2099:
> Yes, when using `use_create_image = true`, pretty sure that's how the feature was designed to work... The "original" root volume is not re-attached. 

Yeah the "original" root device is not re-attached. To illustrate my change with an example, consider this template

```
source "amazon-ebssurrogate" "test" {
  ami_name         = "test_block_device_ami"
  region           = "us-east-1"
  instance_type    = "m3.medium"
  source_ami       = "ami-76b2a71e"
  ssh_username     = "ubuntu"
  use_create_image = true
  # VOL - 1
  launch_block_device_mappings {
    device_name           = "/dev/sda1"
    delete_on_termination = true
    volume_size           = 8
    volume_type           = "gp2"
  }
  # VOL - 2
  launch_block_device_mappings {
    volume_type           = "gp2"
    device_name           = "/dev/xvdb"
    delete_on_termination = true
    volume_size           = 10
  }
  ami_virtualization_type = "hvm"
  ami_root_device {
    source_device_name    = "/dev/xvdb"
    device_name           = "/dev/sda1"
    volume_size           = 10
    volume_type           = "gp2"
  }
}

build {
  sources = ["amazon-ebssurrogate.test"]
}
```

In this example, VOL-1 (`/dev/sda1`) is the initial root volume defined in `launch_block_device_mappings`, and VOL-2 (`/dev/xvdb`) is an additional EBS volume attached during instance launch. When `use_create_image = true`, the plugin detaches both VOL-1 and VOL-2 (since `device_name == ami_root_device.source_device_name`) from the instance, then re-attaches VOL-2 as the new root device. We then create the AMI

With the changes introduced in this PR, the `delete_on_termination` attribute specified for VOL-2 in `launch_block_device_mappings` is now preserved during this re-attachment process. Since VOL-1 is not re-attached to the instance post-detachment, it won't be deleted if the instance is terminated and it must be cleaned up externally.

Let me know if this makes sense to you and if it would address the problem that was raised at [521](https://github.com/hashicorp/packer-plugin-amazon/issues/521)

lorengordon:
@kp2099 Yes, that makes sense, I think we both have the same understanding of the workflow. Though I would want to suggest that the volumes _both_ respect the `delete_on_termination` setting from the config. If they both set it to `true`, then both volumes should be cleaned up by packer (by the delete instance call for the "new" root volume, or by a separate delete volume call for the original root volume).

When we submitted the work for the [PR originally](https://github.com/hashicorp/packer-plugin-amazon/pull/455) (I worked with @dwc0011 on the feature), I'd claim that was the intent, but we missed the cleanup steps.

kp2099:
Hi @lorengordon, before we close out this thread, I wanted to clarify the reasoning behind the decision to detach and re-attach the volume. I reviewed the original PR and saw there was an offline discussion involving Wilken and Lucas from our team.

For the sake of clarity and future reference, it would be helpful to document the original intent behind this change and the rationale that led to it. Could you please share some context on that?

Thanks!

lorengordon:
> Hi @lorengordon, before we close out this thread, I wanted to clarify the reasoning behind the decision to detach and re-attach the volume. I reviewed the original PR and saw there was an offline discussion involving Wilken and Lucas from our team.
> 
> For the sake of clarity and future reference, it would be helpful to document the original intent behind this change and the rationale that led to it. Could you please share some context on that?

The context for why to use the detach and re-attach approach is covered under the original issue: https://github.com/hashicorp/packer-plugin-amazon/issues/440. But essentially, we need to be able to configure the partitions for the root volume, and also support images from the marketplace that have a billing product code for license reasons (like Red Hat Enterprise Linux). Such marketplace images cannot use the snapshot mechanism of RegisterImage, because the billing product code is not maintained with that method. Using CreateImage does maintain the billing product code, but that targets the ec2 instance to create the image, not a snapshot. So we need to configure the ec2 instance to "look" the way we want so CreateImage will capture the correct root volume.

The only real question in the offline discussion was around the naming of the argument, in order to attempt to minimize any user confusion. It was decided that the "best" approach would be to simply name the argument for the api call, CreateImage. Hence we arrived at `use_create_image`. 

(Our original approach named argument `image_method` with a default value of `register` that matched prior RegisterImage behavior, and second supported value of `create` for the CreateImage behavior.)

PR #566, opened 2025-07-10
CRT migration changes

kp2099:


kp2099:
> you also need to remove some of the files which were used to release in github.com

I was planning to do that in a separate PR, but i can remove them in this PR itself.

kp2099:
> > you also need to remove some of the files which were used to release in github.com
> 
> I was planning to do that in a separate PR, but i can remove them in this PR itself.

Removed them

PR #567, opened 2025-07-10
fix: use the aws http client

mogrogan:
----

### Description

Not using the aws client might create problem regarding the imds protocol.

### Resolved Issues
Too soon to know if it's going to fix issue at this moment 

### Rollback Plan

Are there any changes to security controls (access controls, encryption, logging) in this pull request? If so, explain.


common/access_config.go
```diff
@@ -284,14 +283,18 @@ func (c *AccessConfig) getBaseAwsConfig(ctx context.Context) (aws.Config, error)
 	if c.SkipCredsValidation {
 		imdsEnabledState = imds.ClientDisabled
 	}
-	httpClient := cleanhttp.DefaultClient()
-	transport := httpClient.Transport.(*http.Transport)
-	if c.InsecureSkipTLSVerify {
-		transport.TLSClientConfig = &tls.Config{
-			InsecureSkipVerify: true,
-		}
+	httpClient, ok := awshttp.NewBuildableClient().
```
kp2099, line 286: As discussed, I think if we don't build the client here, `GetAwsConfig() `method internally creates a defaultHttp client,
which is an AWS buildable client
ref: 

1. https://github.com/hashicorp/aws-sdk-go-base/blob/ea3a2e071e4deba79032a8d2eb43c2de47ddde75/http_client.go#L11
2. https://github.com/hashicorp/aws-sdk-go-base/blob/ea3a2e071e4deba79032a8d2eb43c2de47ddde75/aws_config.go#L353

let me know what you think.

PR #568, opened 2025-07-11
fix: IMDS credential detection in AMI datasource

### LINKED ISSUE #563: 1.3.8 causes Authentication Issue

#### Overview of the Issue

Running 1.3.8 version throwing error: 
```Error: Datasource.Execute failed: No valid credential sources found\n\n  on packer_template.pkr.hcl line 71:\n  (source code not available)\n\n```

This isn't happening on 1.3.7

#### Reproduction Steps

Steps to reproduce this issue
Use latest plugin version (1.3.8)
### Plugin and Packer version
1.3.8
From `packer version`

OS, Architecture, and any other information you can provide about the
environment.

Include appropriate log fragments. If the log is longer than a few dozen lines,
please include the URL to the [gist](https://gist.github.com/) of the log or
use the [Github detailed format](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d) instead of posting it directly in the issue.

Set the env var `PACKER_LOG=1` for maximum log detail.

jreslock:
Fixes #563

The AMI datasource was failing to detect AWS credentials when using IMDS (Instance Metadata Service) after the migration from builder/common to common in commit 6aada6bc.

The issue was in common/access_config.go where the getBaseAwsConfig method incorrectly used SkipCredsValidation to control IMDS availability instead of SkipMetadataApiCheck.

This fix:
- Changes the IMDS control logic to use SkipMetadataApiCheck instead of SkipCredsValidation
- Ensures IMDS is enabled by default for credential resolution
- Only disables IMDS when skip_metadata_api_check is explicitly set to true
- Adds comprehensive tests to prevent regression

The fix restores the expected behavior where EC2 instances with IAM roles can use IMDS for credential resolution with the AMI datasource.

### DIFF
--- a/common/access_config.go
+++ b/common/access_config.go
@@ -281,7 +281,7 @@ func (c *AccessConfig) Config(ctx context.Context) (*aws.Config, error) {
 
 func (c *AccessConfig) getBaseAwsConfig(ctx context.Context) (aws.Config, error) {
 	imdsEnabledState := imds.ClientEnabled
-	if c.SkipCredsValidation {
+	if c.SkipMetadataApiCheck {
 		imdsEnabledState = imds.ClientDisabled
 	}
 	httpClient := cleanhttp.DefaultClient()
--- a/common/access_config_test.go
+++ b/common/access_config_test.go
@@ -65,3 +65,53 @@ func TestAccessConfigPrepare_UnknownPackerCoreVersion(t *testing.T) {
 		t.Fatalf("packer core version should be unknown, but got %s", c.packerConfig.PackerCoreVersion)
 	}
 }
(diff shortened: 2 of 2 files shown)


mogrogan (approved):
Good catch, did you happen to reproduce the problem that is mentioned in the issue ?

jreslock:
> Good catch, did you happen to reproduce the problem that is mentioned in the issue ?

Hi! Thanks for the approval. I did reproduce using 1.3.8, reverted to 1.3.7 and was not able to reproduce (expected). I made sure that all
of the tests pass before submitting.

PR #573, opened 2025-07-21
Postprocessor: Amazon AMI Import Aws Sdk v2 upgrade

kp2099:
>[!NOTE]
>This PR upgrades the AMI Import Post-Processor to use AWS SDK for Go v2. The goal is to maintain the same functionality while aligning with the newer SDK.

## Changes

1. Migrated from AWS SDK for Go v1 to v2.
2. Polling Behavior Update – In v2, polling for resources is based on `MaxWaitTime`, `MinDelay`, and `MaxDelay`. To preserve existing behavior, `MaxWaitTime` is now derived as:
    ```
    MaxWaitTime = (Max Number of Attempts) * (Polling Delay)
    ```
    This ensures backward compatibility with user-provided settings.
3. Replaced the deprecated `ui.Message()` method with `ui.Say()`.
4. Minor improvements to variable naming for clarity and consistency.


common/state.go
```diff
@@ -123,3 +148,178 @@ func (w *AWSPollingConfig) LogEnvOverrideWarnings() {
...
+	}
+
+	err := waiter.Wait(ctx, imageInput, maxWaitTime, waiterOpts...)
+
+	if err != nil {
+		if strings.Contains(err.Error(), "waiter state transitioned to Failure") {
+			err = fmt.Errorf("failed with ResourceNotReady error, which can "+
+				"have a variety of causes. For help troubleshooting, check "+
+				"our docs: "+
+				"https://www.packer.io/docs/builders/amazon.html#resourcenotready-error\n"+
```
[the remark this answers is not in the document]
kp2099: replaced the url

common/artifact.go
```diff
@@ -0,0 +1,179 @@
...
+		return images
+	}
+
+	data, ok := a.StateData["generated_data"].(map[string]interface{})
+	if !ok {
+		return images
+	}
+
+	for _, image := range images {
+		image.SourceImageID = data["SourceAMI"].(string)
```
anurag5sh, line 175: Will the `SourceAMI` always be available in the data? And is it always a string? If not, we may need to add checks or handle errors.

post-processor/import/post-processor.go
```diff
@@ -291,78 +300,79 @@ func (p *PostProcessor) PostProcess(ctx context.Context, ui packersdk.Ui, artifa
...
+		importStart, err = client.ImportImage(ctx, params)
 		return err
 	})
 
 	if err != nil {
 		return nil, false, false, fmt.Errorf("Failed to start import from s3://%s/%s: %s", p.config.S3Bucket, p.config.S3Key, err)
 	}
 
-	ui.Message(fmt.Sprintf("Started import of s3://%s/%s, task id %s", p.config.S3Bucket, p.config.S3Key, *import_start.ImportTaskId))
+	ui.Message(fmt.Sprintf("Started import of s3://%s/%s, task id %s", p.config.S3Bucket, p.config.S3Key, *importStart.ImportTaskId))
```
anurag5sh, line 324: We can replace ui.Message with ui.Say since Message is deprecated.

common/boot_mode_validation.go
```diff
@@ -0,0 +1,20 @@
...
+
+package common
+
+import (
+	"fmt"
+	"slices"
+)
+
+// IsValidBootMode checks that the bootmode is a value supported by AWS
+func IsValidBootMode(bootmode string) error {
```
mogrogan, line 12: nit: names looks like a function that return a boolean (feel free to ignore)

kp2099: hi martin, i believe i kept the fn name as is, as it was introduced in v1, we can refactor the name in a subsequent PR.

common/state.go
```diff
@@ -123,3 +148,185 @@ func (w *AWSPollingConfig) LogEnvOverrideWarnings() {
...
+		envOverrides.awsMaxAttempts.Val = w.MaxAttempts
+		envOverrides.awsMaxAttempts.overridden = true
+	}
+	if w.DelaySeconds != 0 {
+		envOverrides.awsPollDelaySeconds.Val = w.DelaySeconds
+		envOverrides.awsPollDelaySeconds.overridden = true
+	}
+
+	waitOpts := applyEnvOverrides(envOverrides)
+	return waitOpts
```
mogrogan, line 226: nit: could just 
```
return applyEnvOverrides(envOverrides)
```

mogrogan (approved):
I think this is good ! 

Great work !

common/state.go
kp2099: i will address it in the subsequent PR.

Issue #599, opened 2025-09-02
Regression in 1.4.0: Error: Datasource.Execute failed: validating provider credentials: retrieving caller identity from STS: already exists, tfUserAgentAppender


mjnowen:
#### Overview of the Issue

This is not an issue in <= v1.3.10 plugin and only in v1.4.0 and onwards.

```
data "amazon-parameterstore" "base-ami" {
  name   = "/aws/service/canonical/ubuntu/server/24.04/stable/current/amd64/hvm/ebs-gp3/ami-id"
}
```

will fail with this error message:

```
Error: Datasource.Execute failed: validating provider credentials: retrieving caller identity from STS: already exists, tfUserAgentAppender
```

if the environment has this ENV VAR: `TF_APPEND_USER_AGENT` present when you execute packer.

ChatGPT says:

> Cause: A bug in the amazon plugin v1.4.0 double-registers an AWS SDK user-agent middleware (“tfUserAgentAppender”) during STS GetCallerIdentity, often when TF_APPEND_USER_AGENT is present in the environment. That collision surfaces as “already exists, tfUserAgentAppender”.

kp2099:
Hi @mjnowen, I’ve opened a PR that should address this issue. Could you give it a try and let me know if it resolves the problem on your side?
PR: https://github.com/hashicorp/packer-plugin-amazon/pull/600

mjnowen:
This is great news and thanks for jumping on this. How can I 'test' this fix without a new release being created? If you can provide instructions, I can test.

kp2099:
Hi @mjnowen in the template file can you set the version as
```
version = "= 1.5.0"
```
this should let you run it.

mjnowen:
pre-release is still being ignored

```
ubuntu@local:~/image$ PACKER_LOG=1 packer init .
2025/09/08 11:52:02 [INFO] Packer version: 1.14.1 [go1.23.2 linux amd64]
2025/09/08 11:52:02 [INFO] PACKER_CONFIG env var not set; checking the default config file path
2025/09/08 11:52:02 [INFO] PACKER_CONFIG env var set; attempting to open config file: /home/ubuntu/.packerconfig
2025/09/08 11:52:02 [WARN] Config file doesn't exist: /home/ubuntu/.packerconfig
2025/09/08 11:52:02 Old default config directory found: /home/ubuntu/.packer.d
2025/09/08 11:52:02 Old default config directory found: /home/ubuntu/.packer.d
[...]

PR #600, opened 2025-09-03
fix UserAgentProducts when building the request.

kp2099:
Currently when building the config, we are passing UserAgentProducts as part of the [Config](https://github.com/hashicorp/aws-sdk-go-base/blob/main/internal/config/config.go#L73) struct. But if `TF_APPEND_USER_AGENT` environment variable is set, this causes a collision as both the `UserAgentProducts` and the env var were using the same ID in the middleware Stack BuildStep. This collision was causing the build to fail.

Also  default User-Agent middleware prepends itself to the contents of the User-Agent header, we have to prepend our custom User-Agent. This happens if we set the UserAgentProducts in the `APNInfo`

Closes: https://github.com/hashicorp/packer-plugin-amazon/issues/599

Screenshot of how the user agent will look like

Screenshot 2025-09-03 at 5 14 56 PM

### DIFF
--- a/common/access_config.go
+++ b/common/access_config.go
@@ -282,8 +282,6 @@ func (c *AccessConfig) getBaseAwsConfig(ctx context.Context) (aws.Config, error)
 		imdsEnabledState = imds.ClientDisabled
 	}
 	userAgentProducts := awsbase.UserAgentProducts{
-		{Name: "APN", Version: "1.0"},
-		{Name: "HashiCorp", Version: "1.0"},
 		{Name: "packer-plugin-amazon", Version: pluginversion.Version, Comment: "+https://www.packer.io/docs/builders/amazon"},
 	}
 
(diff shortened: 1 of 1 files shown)


common/access_config.go
```diff
@@ -282,17 +282,20 @@ func (c *AccessConfig) getBaseAwsConfig(ctx context.Context) (aws.Config, error)
 		imdsEnabledState = imds.ClientDisabled
 	}
 	userAgentProducts := awsbase.UserAgentProducts{
-		{Name: "APN", Version: "1.0"},
-		{Name: "HashiCorp", Version: "1.0"},
 		{Name: "packer-plugin-amazon", Version: pluginversion.Version, Comment: "+https://www.packer.io/docs/builders/amazon"},
```
kp2099, line 285: i can probably change the URL to the latest one here

PR #639, opened 2025-12-02
fix: improve instance profile propagation handling

afiguera:
### Description

* Our github actions workflows do bakes of AMIs using spot instances. After upgrading the plugin to the 1.8.0 version,  bakes started failing. When doing local testing of the plugin, it reveals that the existing 5-second delay is insufficient for IAM instance profile propagation to EC2. In all the cases (both locally and github actions), spot fleet creation consistently fails with "Invalid IAM Instance Profile name" and times out. Increasing the delay to 7-8 seconds resolves the issue, but this approach is unreliable as propagation times can vary based on AWS region, load, or other factors.

#### Fix Proposal

* When creating the instance profile, replace the fixed 5-second sleep with proper API-based validation to ensure EC2 can see IAM instance profile changes before proceeding to spot instance fleet creation. This eliminates the guesswork of IAM eventual consistency timing.

### DIFF
--- a/common/step_iam_instance_profile.go
+++ b/common/step_iam_instance_profile.go
@@ -175,7 +175,7 @@ func (s *StepIamInstanceProfile) Run(ctx context.Context, state multistep.StateB
 		}
 
 		// In aws sdk go v2, we noticed if there was no Wait, the spot fleet requests were failing even with retry.
-		// Running dummy instance in DryRun mode to validate that the instance profile is visible to EC2
+		// Running a dummy instance in DryRun mode to validate that the instance profile is visible to EC2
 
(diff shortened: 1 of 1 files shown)


common/step_run_spot_instance.go
```diff
@@ -420,12 +420,6 @@ func (s *StepRunSpotInstance) Run(ctx context.Context, state multistep.StateBag)
 	err = retry.Config{
 		Tries: 11,
 		ShouldRetry: func(err error) bool {
-			if strings.Contains(err.Error(), "Invalid IAM Instance Profile name") {
```
kp2099: Just to verify: did we remove this check because StepInstanceProfile now handles capturing and retrying this error?

kp2099 (approved):

PR #654, opened 2026-03-18
add elastic_ip_allocation_id support to EBS builders

### LINKED ISSUE #558: Reserved Private IP for AWS Instance in Packer.
Hi
I want to use a reserved private IP for an AWS instance in Packer with/without the "associate_public_ip_address" se to false. In GCP, there's a parameter for this - "**network_ip**".
Is there any similar parameter in AWS that can serve the same purpose?

Please redirect me if there's another space that I need to post this query in.

Thanks in advance.

tjnicholas:
### Description

Introduces a new `elastic_ip_allocation_id` string field on RunConfig that allows users to associate a pre-existing Elastic IP with the Packer build instance immediately after launch. 

This is useful when the build host, or other build time dependency, needs to be reachable from a known, stable public IP — for example, to satisfy firewall allowlists without relying on an auto-assigned address.

Implementation details:
- Adds `StepAssociateEIP` to the multistep pipeline in both the `ebs` and `ebssurrogate` builders, positioned after instance launch and before password/SSH steps so the EIP is in place before any connection is attempted.
- On Run, the step calls AssociateAddress then re-describes the instance and writes the refreshed object back to the state bag, ensuring SSHHost picks up the EIP as PublicIpAddress.
- On Cleanup, the step calls DisassociateAddress (but does not release the EIP — it belongs to the caller).
- Extends the Ec2Client interface with AssociateAddress and DisassociateAddress; the real *ec2.Client already satisfies both.
- Validates that the value matches `^eipalloc-[0-9a-f]+$` and that it is not combined with associate_public_ip_address, which would be redundant and ambiguous.

### Resolved Issues

This addresses the AWS portion of #36 , but not the Azure portion. 
Closes #558, or rather enables a useful answer. 

### Rollback Plan

### DIFF
--- a/builder/ebs/builder.go
+++ b/builder/ebs/builder.go
@@ -360,6 +360,9 @@ func (b *Builder) Run(ctx context.Context, ui packersdk.Ui, hook packersdk.Hook)
 			LaunchMappings: b.config.LaunchMappings,
 		},
 		instanceStep,
+		&awscommon.StepAssociateEIP{
+			AllocationId: b.config.ElasticIpAllocationId,
+		},
 		&awscommon.StepGetPassword{
 			Debug:     b.config.PackerDebug,
 			Comm:      &b.config.RunConfig.Comm,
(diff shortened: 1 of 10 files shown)


tjnicholas:
Anything I can add to this to make merging more attractive? I wonder, for example, if it would be better to be able to specify a list of eip IDs. That would make it easier to have multiple packer runs happen in parallel.

PR #683, opened 2026-06-15
Make EBS source termination wait optional

### LINKED ISSUE #682: EBS source instance termination wait can dominate cleanup
### Overview

I think some Amazon EBS workflows could benefit from an explicit option to return after EC2 accepts source instance termination.

The case I measured used Packer to run work on a source instance and collect outputs before cleanup. The build did not need Packer to wait for the instance to reach the final `terminated` state.

### Measurements

In a direct EC2 lifecycle comparison, `c5.large` reached `terminated` in about one minute after `TerminateInstances` was accepted. `c5.metal` remained `shutting-down` for more than 19 minutes before later reaching `terminated`.

In a minimal local Packer run on `c5.metal`, EC2 accepted source instance termination around one second after Packer requested it. Waiting for the final state was the long part.

### Expected behavior

My theory is that the current default wait should stay unchanged. A separate opt in setting would let callers choose the faster completion point when their workflow does not need Packer to wait for final source instance deletion.

### Scope

This is separate from `skip_create_ami`. The option should not require image creation to be disabled. It is about the cleanup completion point after EC2 accepts source instance termination.

FocalChord:
### Description

This PR adds an Amazon EBS builder option named `skip_source_instance_termination_wait`.

When the option is false, behavior is unchanged. Packer still waits for EC2 to report the source instance as `terminated`.

When the option is true, Packer returns from source instance cleanup after `TerminateInstances` succeeds. This keeps the choice explicit and separate from `skip_create_ami`.

This PR does not change AMI creation or AMI preparation behavior.

### Resolved Issues

Closes #682

### Evidence

The direct EC2 data I saw suggests the final source instance state can take much longer on bare metal instances than the API acceptance of termination.

1. `c5.large` reached `terminated` in about one minute after `TerminateInstances` was accepted.
2. `c5.metal` remained `shutting-down` for more than 19 minutes before later reaching `terminated`.
3. A minimal local Packer run on `c5.metal` saw EC2 accept source instance termination around one second after Packer requested it.

### Validation

1. `go test ./builder/common ./common ./builder/ebs`
2. Local plugin build completed
3. Git whitespace check passed

### Rollback Plan

Revert this change.

No. This does not change access controls, encryption, or logging.


FocalChord:
Hi, I signed the CLA but unsure why it hasn't gone through yet.

PR #700, opened 2026-08-21
fix: scope IAM instance profile propagation check to IAM only

### LINKED ISSUE #675: v1.8.1 regression: amazon-ebs fails in non-default-VPC accounts when using temporary_iam_instance_profile_policy_document
### Overview of the Issue
amazon-ebs builds that work with packer-plugin-amazon v1.8.0 fail with v1.8.1 in AWS accounts/regions that do not have a default VPC.

The failing configuration uses:
- temporary_iam_instance_profile_policy_document
- subnet_filter
- the temporary security group path

### Regression window:
- works with v1.8.0
- fails with v1.8.1
- reverting to v1.8.0 resolves the issue with no template changes

The failure appears to happen in the new IAM instance profile propagation validation added in v1.8.1. The new RunInstances(DryRun) check seems to run without the subnet/VPC/security group context used by the actual launch. In accounts without a default VPC, that returns VPCIdNotSpecified.

There is also a secondary cleanup problem after the failure: the temporary role remains attached to the temporary instance profile, so cleanup hits DeleteConflict on DeleteRole and DeleteInstanceProfile.

### Reproduction Steps
1. Use packer-plugin-amazon v1.8.1.
2. Run an amazon-ebs build in an AWS account/region with no default VPC.
3. Use a builder config that includes:
  - temporary_iam_instance_profile_policy_document
  - subnet_filter
  - temporary security group creation
4. Start the build.
5. Observe failure during Waiting for the change to propagate because of eventual consistency...
6. Re-run the same build with only the plugin version changed back to v1.8.0.
7. Observe that the build succeeds.

### Plugin and Packer version
packer-plugin-amazon v1.8.1

### Regression window:
packer-plugin-amazon v1.8.0  -> works
packer-plugin-amazon v1.8.1  -> fails

```
packer {
  required_plugins {
    amazon = {
      version = "1.8.1"
    }
  }
}
source "amazon-ebs" "example" {
  region                            = var.aws_region
  instance_type                     = "t3.small"
  source_ami                        = var.source_ami
  ssh_username                      = "ubuntu"
  temporary_security_group_source_cidrs = ["10.194.80.0/23"]
  temporary_iam_instance_profile_policy_document {
    Statement {
      Action   = ["ssm:GetParameter"]
      Effect   = "Allow"
      Resource = ["*"]
    }
    Version = "2012-10-17"
  }
  subnet_filter {
    filters = {
      "owner-id" = var.aws_account
    }
    most_free = true
    random    = false
  }
}
build {
  sources = ["source.amazon-ebs.example"]
}
```

- CI runner environment
- amazon-ebs builder
- AWS account/region with no default VPC
- launch path uses subnet_filter
- temporary IAM instance profile path in use
- reproduced on both amd64 and arm64 builds
Primary failure:
==> amazon-ebs.noble-amd64: Waiting for the change to propagate because of eventual consistency...
Build 'amazon-ebs.noble-amd64' errored after 11 seconds 663 milliseconds: timed out waiting for IAM changes to propagate to EC2: operation error EC2: RunInstances, https response error StatusCode: 400, RequestID: c5748273-4a36-4da1-ae26-fb9c0d4d9796, api error VPCIdNotSpecified: No default VPC for this user. GroupName is only supported for EC2-Classic and default VPC.
Secondary cleanup failure after that:
==> amazon-ebs.noble-amd64: Error operation error IAM: DeleteRole, https response error StatusCode: 409, RequestID: c2f85e34-7269-45a5-9b7d-6b8f7f818362, DeleteConflict: Cannot delete entity, must remove roles from instance profile first.. Please delete the role manually: packer-...
==> amazon-ebs.noble-amd64: Error operation error IAM: DeleteInstanceProfile, https response error StatusCode: 409, RequestID: 697ab8a3-1159-48b6-bef4-d445b69e93f8, DeleteConflict: Cannot delete entity, must remove roles from instance profile first.. Please delete the instance profile manually: packer-...

### Expected behavior:
- Builds that launch into a non-default VPC via subnet_filter should continue to work in v1.8.1.
- The IAM propagation validation should use the same launch networking context as the real instance launch, or avoid requiring a default VPC.
- Cleanup should detach the role from the instance profile even if the validation step fails.
Suspected regression point:
- v1.8.1 release note: Fix instance profile propagation handling
- The new RunInstances(DryRun) validation appears to omit subnet/VPC/security-group context.

cchristous:
### Description

Fixes a v1.8.1 regression (introduced in #639) in the temporary IAM instance profile flow.

#639 replaced the post-attach `time.Sleep` with a `RunInstances` dry-run that waits for the newly-created instance profile to propagate to EC2. However, the dry-run hardcoded `InstanceType: t3.nano` and passed no networking context, so it could fail for reasons unrelated to IAM propagation:

- **non-x86_64 (e.g. arm64) source AMIs** → architecture mismatch, since `t3.nano` is x86_64.
- **accounts/regions without a default VPC** → `VPCIdNotSpecified` (with no subnet, EC2 falls back to the default VPC and resolves a security group by name).

`ShouldRetry` only matched `Invalid IAM Instance Profile`, so both of these were treated as terminal and the build failed at the propagation check — even though the real launch (which uses the configured instance type and the resolved subnet/security groups) would have succeeded. Separately, `roleIsAttached` was only set *after* the dry-run passed, so a failed dry-run left the temporary role attached to the instance profile, producing `DeleteConflict` on `DeleteRole`/`DeleteInstanceProfile` during cleanup.

**Change.** The propagation check has one job: confirm the instance profile is visible to EC2. This scopes it to exactly that:

- The dry-run now treats any non-`Invalid IAM Instance Profile` outcome — `DryRunOperation` *or* a downstream error such as an architecture or instance-type/AZ mismatch — as "EC2 got past IAM validation, so the profile has propagated." Everything else is deferred to the real launch step, which validates and reports it with accurate context.
- To keep the request well-formed enough to reach the IAM check (rather than being rejected earlier for an invalid type or the default-VPC fallback), it now passes the configured instance type and the resolved subnet. Instance-type resolution moved to a `RunConfig.EffectiveInstanceType()` helper that falls back to the first `spot_instance_types` entry when `instance_type` is unset (spot-only builds), since `Prepare` guarantees exactly one is set.
- `roleIsAttached` is set as soon as the role is attached, so `Cleanup` detaches it even if the propagation check fails.
- On failure, the error message now distinguishes an actual propagation timeout from an unrelated request error, instead of always reporting "timed out waiting for IAM changes to propagate."

This affects the `amazon-ebs`, `amazon-ebsvolume`, and `amazon-ebssurrogate` builders (which share this step). The `amazon-instance` builder is unaffected — it uses a separate step that never had the dry-run.

**Alternatives considered.** Making the dry-run fully mirror the real launch (subnet + security groups + exact type) was rejected: security groups add no signal for an IAM-propagation check and introduce a fresh eventual-consistency surface (a just-created temporary security group can transiently return `InvalidGroup.NotFound`), and pinning a single instance type is stricter than a spot fleet, which launches on any of several types available in the AZ. Scoping the check to IAM alone is simpler and defers every non-IAM error to the real launch rather than treating it as a propagation failure, so it stops failing the builds the old code broke.

**Testing.** `go build ./...`, `go vet`, and the `common`, `builder/ebs`, `builder/ebsvolume`, and `builder/ebssurrogate` package tests pass. Added unit tests for `EffectiveInstanceType()` (including the spot fallback) and for the dry-run request construction.

### Resolved Issues

Closes #675

### Rollback Plan

No changes to access controls, encryption, or logging. The temporary IAM role/instance-profile lifecycle and the policy documents applied are unchanged; the granted permissions are identical. The only security-adjacent effect is a reliability improvement to cleanup: the temporary role is now reliably detached and deleted even when the propagation check fails, avoiding leftover temporary roles/instance profiles.


hariom-hashicorp:
Thanks for the fix, @cchristous! I reviewed and tested it.

* Reproduced the ARM64 issue on v1.8.1
* Verified the changes and unit test coverage
* Confirmed we now use the configured instance type instead of the hardcoded `t3.nano`
* Confirmed the resolved subnet is passed through, which fixes the no-default-VPC case
* The cleanup issue is also fixed by setting `roleIsAttached` before the check

**One non-blocking question:**
The check for a missing `subnet_id` makes sense as a safeguard. Do you know if there are any cases where `subnet_id` could actually be missing at this point? My understanding is that `StepNetworkInfo` should always set it before this step, but I may be missing an edge case.

Other than that, everything looks good.

cchristous:
Thanks for testing it, @hariom-hashicorp! Your understanding is correct — there's no reachable path today where `subnet_id` is missing. `StepNetworkInfo` runs before this step in all three builders (`amazon-ebs`/`ebsvolume`/`ebssurrogate`) and always calls `state.Put("subnet_id", …)`. It can be an empty string (when no `subnet_id`/`subnet_filter` is set and a default VPC is used), which the dry-run request handles by simply omitting `SubnetId` — but the key is always present.

The guard is purely defensive/consistency: the line previously used a bare `state.Get("subnet_id").(string)`, which would panic on a nil interface if the key were ever absent, so I switched it to the same `GetOk` + clean-error form already used for `source_image` a few lines above. That way, if this shared step is ever wired into a flow without `StepNetworkInfo` ahead of it, it fails with an actionable error instead of panicking. Happy to drop it if you'd prefer to keep the step minimal — it's not guarding a currently-reachable case.

hariom-hashicorp (approved):
LGTM

